PR #1, opened 2020-02-07
W18 Viktor Dahl

tehvicke:
Link to the project is in the readme

### DIFF
--- a/server.js
+++ b/server.js
@@ -1,21 +1,21 @@
-import express from "express";
-import bodyParser from "body-parser";
-import cors from "cors";
-import mongoose from "mongoose";
+import express from 'express'
+import bodyParser from 'body-parser'
+import cors from 'cors'
+import mongoose from 'mongoose'
 
-const MAXPERPAGE = 10; /* max results per page */
+const MAXPERPAGE = 10 /* max results per page */
 
 // import goldenGlobesData from './data/golden-globes.json'
-import avocadoSalesData from "./data/avocado-sales.json";
+import avocadoSalesData from './data/avocado-sales.json'
 // import booksData from './data/books.json'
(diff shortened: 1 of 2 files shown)


server.js
```diff
@@ -27,9 +51,99 @@ const app = express()
...
+  const dbQuery = queryBuilder(req, res)
+
+  const numberOfPages = Math.ceil(
+    (await AvocadoSales.countDocuments(dbQuery)) / MAXPERPAGE
+  ) /* Not really needed to send for each new page request - could store in frontend and pass bool to not run this function */
+
+  if (numberOfPages != 0 && numberOfPages < page) {
+    /* If page provided is too high return last page */
+    page = numberOfPages
+  }
```
puj: This is really clever! I like this!
Try `page=0` ;)

server.js
```diff
@@ -27,9 +51,99 @@ const app = express()
...
+  })
+})
+
+app.get('/avocado-sales/:id', async (req, res) => {
+  
+  try {
+    const result = await AvocadoSales.findOne({ _id: req.params.id })
+    if (result === null) {
+      throw 'error'
+    }
```
puj: Usually if you `throw` inside of a `catch` you should provide a more descriptive message for your catch statement, and then use that information to populate your response.

server.js
```diff
@@ -27,9 +51,99 @@ const app = express()
...
+      status: 'failure',
+      message: 'Please ensure you pick two dates'
+    })
+  }
+  /* Add mandatory fields to the query */
+  const dbQuery = {
+    date: {
+      /* check for date range greater or equal && lesser than */
+      $gte: new Date(start),
+      $lt: new Date(end)
```
puj: Good job leveraging mongo

PR #2, opened 2020-02-08
Project 18 - Sofie Nyblad

nyblad:
**Live project**
Backend with examples of endpoints:
https://nyblad-guest-list.herokuapp.com/guests
https://nyblad-guest-list.herokuapp.com/guests?attending=true
https://nyblad-guest-list.herokuapp.com/guests?name=nyblad
https://nyblad-guest-list.herokuapp.com/guests/5e3f1f58bbf85b00237c9ded

Frontend: https://the-nyblad-wedding.netlify.com/

**Summary of the week**
Was a bit hard to get it mongoosy, but with some help and googling it works as I wanted. At first I tried to do pagination in backend to but since I didn't get the result I wanted there I did it in the frondend instead. 

**Question**
- I can search for first or last name ("sofie" or "nyblad"), but not both at the same time ("sofie nyblad") - how to do that? Suppose the easy way to go is to have first and last name in one filed in the model instead of separate?
- When I want to filter on name or attending in the frontend, should the queries be handled with dispatches to the reducer or can I handle it localy in the file I'm mapping the guests? Feels like I'm complicating stuff: https://github.com/nyblad/nyblad-final-project/blob/master/src/components/GuestList.js


models/guest.js
```diff
@@ -0,0 +1,13 @@
...
+  first_name: { type: String },
+  last_name: { type: String },
+  email: { type: String },
+  phone: { type: Number },
+  allergies: { type: String },
+  other: { type: String },
+  added: { type: Date, default: Date.now },
+  updated: { type: Date, default: Date.now },
+  isAttending: { type: Boolean },
+})
```
puj: Awesome that you lifted the model to its own file!

server.js
```diff
@@ -2,36 +2,109 @@ import express from 'express'
...
+  } else {
+    res.status(503).json({ error: 'Service unavailabale' })
+  }
+})
 
-// Start defining your routes here
+// ROUTES FOR GET
 app.get('/', (req, res) => {
-  res.send('Hello world')
+  res.send('Endpoints: GET/guests GET/guests/:id GET/guests?name= GET/guests?attending=true/false')
```
puj: Awesome that you include usage in your default page.  I believe it's `guest-list`.  You could solve this discrepancy by making the name of your endpoint a `const`

server.js
```diff
@@ -2,36 +2,109 @@ import express from 'express'
...
+  }
+})
+
+// GET ROUTE FOR SPECIFIC ID
+app.get('/guests/:id', async (req, res) => {
+  const guest = await Guest.findById(req.params.id)
+  if (guest) {
+    res.json(guest)
+  } else {
+    res.status(404).json({ error: 'Guest not found' })
```
puj: 404 👍

server.js
```diff
@@ -2,36 +2,109 @@ import express from 'express'
...
+// })
+// ROUTES FOR PUT
+// app.put('/guests/:id', (req, res) => {
+//   return res.send(`PUT HTTP method on guest/${req.params.id} resource`)
+// })
+//ROUTES FOR DELETE
+// app.delete('/guests/:id', (req, res) => {
+//   return res.send(`DELETE HTTP method on guest/${req.params.id} resource`)
+// })
+
```
puj: Always nice to leave notes to yourself to save some work in the future

PR #4, opened 2020-02-09
https://globes-details.herokuapp.com/

Sara-Borgstrom:


server.js
```diff
@@ -27,12 +77,34 @@ const app = express()
...
+      res.json(results);
+    }).catch((err) => {
+      res.json({ message: 'Cannot find this film', err: err })
+    });
+});
+
 // Start defining your routes here
 app.get('/', (req, res) => {
-  res.send('Hello world')
+  res.send('End points: /globes/_id/*** or /gloves?q=***')
```
puj: 👍 for usage examples

server.js
```diff
@@ -27,12 +77,34 @@ const app = express()
...
+  const queryString = req.query.q;
+  const queryRegex = new RegExp(queryString, "i")
+  Globe.find({ 'nominee': queryRegex }).then(results => {
+    if (results.length > 0) {
+      res.json(results)
+    } else {
+      res.status(404).json({ error: 'Cannot find this nominee' })
+    }
+  });
+});
```
puj: A nominee is a pretty specific thing to search for.  It would be interesting to think how you could construct an API endpoint called
`app.get('/globes/nominees'...)`
And perhaps use it like 
`http://localhost:9000/globes/nominees?name=Sara`

puj: The idea behind this example is to think that if the client uses `/globes` it might not be clear that you are filtering on nominee name 🤔

PR #5, opened 2020-02-09
Project Mongo API – Emmie Nilsson

eilsson:
Live project
https://mongo-top-music.herokuapp.com/

I felt very unsure about error handling, should I do it some other way?


README.md
```diff
@@ -1,13 +1,22 @@
...
+A RESTful API for top music built with Express and using Mongoose models and MongoDB.
+The API has different routes for tracks and artists and some filtering options.
 
 ## The problem
+The data is taken from a JSON file that contains data about 50 top tracks from Spotify. I began with planning what data I wanted to include from the JSON file and what routes I would create.
 
-Describe how you approached to problem, and what tools and techniques you used to solve it. How did you plan? What technologies did you use? If you had more time, what would be next?
+After that I built a model with Mongoose and a seeding function to populate the database.
+
+I created and tested each route, added queries to tracks, and error messages for when track or artist isn't found.
```
puj: Great description of your process! 👍

server.js
```diff
@@ -3,36 +3,100 @@ import bodyParser from 'body-parser'
...
 
-// If you're using one of our datasets, uncomment the appropriate import below
-// to get started!
-// 
-// import goldenGlobesData from './data/golden-globes.json'
-// import avocadoSalesData from './data/avocado-sales.json'
-// import booksData from './data/books.json'
-// import netflixData from './data/netflix-titles.json'
-// import topMusicData from './data/top-music.json'
+import topMusicData from './data/top-music.json'
```
puj: Good cleanup!

server.js
```diff
@@ -3,36 +3,100 @@ import bodyParser from 'body-parser'
...
 const mongoUrl = process.env.MONGO_URL || "mongodb://localhost/project-mongo"
 mongoose.connect(mongoUrl, { useNewUrlParser: true, useUnifiedTopology: true })
 mongoose.Promise = Promise
 
-// Defines the port the app will run on. Defaults to 8080, but can be 
-// overridden when starting the server. For example:
-//
-//   PORT=9000 npm start
-const port = process.env.PORT || 8080
+const port = process.env.PORT || 8082
```
puj: Why 8082? 😄

server.js
```diff
@@ -3,36 +3,100 @@ import bodyParser from 'body-parser'
...
+      new Track(trackData).save()
+    })
+  }
+  seedDatabase()
+}
+
+app.get('/', async (req, res) => {
+  const tracks = await Track.find()
+  res.json(tracks)
+})
```
puj: Interesting choice to return the tracks here

server.js
```diff
@@ -3,36 +3,100 @@ import bodyParser from 'body-parser'
...
+
+  if (genreQuery && bpmQuery) {
+    tracks = await Track.find({
+      $and: [{ genre: genreQuery }, { bpm: bpmQuery }]
+    })
+  } else if (genreQuery || bpm) {
+    tracks = await Track.find({
+      $or: [{ genre: genreQuery }, { bpm: bpmQuery }]
+    })
+  }
```
puj: Interesting logic.  I wonder what would happen if you included another field here.  How the if statement would look in that case.

server.js
```diff
@@ -3,36 +3,100 @@ import bodyParser from 'body-parser'
...
+    res.json(track)
+  }
+})
+
+app.get('/artists', async (req, res) => {
+  const artists = await Track.find()
+  res.json(artists)
+})
+
+app.get('/artists/:artist/tracks', async (req, res) => {
```
puj: 🏆 REST 🏆

server.js
```diff
@@ -3,36 +3,100 @@ import bodyParser from 'body-parser'
...
+}
+
+app.get('/', async (req, res) => {
+  const tracks = await Track.find()
+  res.json(tracks)
+})
+
+app.get('/tracks', async (req, res) => {
+  const genreQuery = new RegExp(req.query.genre, 'i')
+  const bpmQuery = req.query.bpm
```
puj: Cool you combined two different types in the same filter logic!

server.js
```diff
@@ -3,36 +3,100 @@ import bodyParser from 'body-parser'
...
+    tracks = await Track.find({
+      $or: [{ genre: genreQuery }, { bpm: bpmQuery }]
+    })
+  }
+
+  if (tracks.length > 0) {
+    res.json(tracks)
+  } else {
+    res.status(404).json({ message: "No tracks found" })
+  }
```
puj: Try this
https://mongo-top-music.herokuapp.com/tracks?bpm=pop

server.js
```diff
@@ -3,36 +3,100 @@ import bodyParser from 'body-parser'
...
+  res.json(artists)
+})
+
+app.get('/artists/:artist/tracks', async (req, res) => {
+  const artistName = new RegExp(req.params.artist, 'i')
+  let artist
+  try {
+    artist = await Track.find({ 'artistName': artistName })
+  } catch (error) {
+    return null
```
puj: This should probably be where you send a response back to the client, a good use case for `res.status(400)`

server.js
eilsson: I had forgotten the server on port 8080 for another project, so I had to use another. Forgot to change this back to 8080 later. :)

server.js
eilsson: Yeah, I was unsure about this. Probably overkill to have it when I already return tracks in the tracks endpoint.

server.js
eilsson: I see the problem. Will look into this. Thank you :)

server.js
eilsson: Thank you. You're right. Have changed that now.

PR #6, opened 2020-02-09
Project 18 Mongo API - Jenny Jacobsson

jennyjacobsson:
**Live project:**
https://jenny-project-18-mongo.herokuapp.com/

I'm glad to say that this backend stuff doesn't seem to be as hard and complex as I first thought, even though I've struggled a bit during these past weeks. 😅 

This week I had a problem with the error messages at first. If I searched for a title that didn't exist I just got an empty array back instead of a message. So I got a tip to use throw new error to catch that, and now I'm glad to say it works. :)

### DIFF
--- a/server.js
+++ b/server.js
@@ -56,19 +56,28 @@ const Book = mongoose.model('Book', {
   }
 })
 
-// const addBooksToDataBase = () => {
-//   booksData.forEach((book) => {
-//     new Book(book).save()
-//   })
-// }
+if (process.env.RESET_DB) {
+  const addBooksToDataBase = async () => {
+    await Book.deleteMany({})
 
-// addBooksToDataBase()
+    booksData.forEach((book) => {
+      new Book(book).save()
+    })
+  }
+
+  addBooksToDataBase()
+}
 
 // Add middlewares to enable cors and json body parsing
 app.use(cors())
 app.use(bodyParser.json())
 
 // Start defining your routes here
+
+app.get('/', (req, res) => {
+  res.send('Endpoints: /books (search by title, author or use sort_by=average_rating to get the books sorted), /books/:isbn (search for a single book)')
+})
+
 app.get('/books', (req, res) => {
   const { title, author, sort_by } = req.query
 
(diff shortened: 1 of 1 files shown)


server.js
```diff
@@ -23,15 +23,106 @@ mongoose.Promise = Promise
...
+
 // Add middlewares to enable cors and json body parsing
 app.use(cors())
 app.use(bodyParser.json())
 
 // Start defining your routes here
+
 app.get('/', (req, res) => {
-  res.send('Hello world')
+  res.send('Endpoints: /books (search by title, author or use sort_by=average_rating to get the books sorted), /books/:isbn (search for a single book)')
```
puj: 👍

server.js
```diff
@@ -23,15 +23,106 @@ mongoose.Promise = Promise
...
 })
 
+app.get('/books', (req, res) => {
+  const { title, author, sort_by } = req.query
```
puj: Nice destructuring!

server.js
```diff
@@ -23,15 +23,106 @@ mongoose.Promise = Promise
...
+
+  // Check if sort by is a descending sortable
+  if (sort_by && ['average_rating'].includes(sort_by)) {
+    sort[sort_by] = -1
+  }
+
+  // Check if sort by is a ascending sortable
+  if (sort_by && ['authors'].includes(sort_by)) {
+    sort[sort_by] = 1
+  }
```
puj: Very cool! Allowing the client to choose the `sort_by`

server.js
```diff
@@ -23,15 +23,106 @@ mongoose.Promise = Promise
...
+  res.send('Endpoints: /books (search by title, author or use sort_by=average_rating to get the books sorted), /books/:isbn (search for a single book)')
 })
 
+app.get('/books', (req, res) => {
+  const { title, author, sort_by } = req.query
+
+  const sort = {}
+
+  // Check if sort by is a descending sortable
+  if (sort_by && ['average_rating'].includes(sort_by)) {
```
puj: Why `['average_rating'].includes(sort_by)` instead of `'average_rating' === sort_by`?

server.js
```diff
@@ -23,15 +23,106 @@ mongoose.Promise = Promise
...
+    .sort(sort)
+    .then((results) => {
+      if (results.length === 0) {
+        throw new Error('No results found')
+      }
+      res.json(results)
+    }).catch((err) => {
+      res.json({ message: err.message })
+    })
+})
```
puj: Clean

PR #9, opened 2020-02-09
Mongo-project

elol42:
https://api-mongoose-project.herokuapp.com/

A question that a have is when i use error, when I use a promise as I did in my code do I need to have if status === null to get error message. If a use async and await can I do a error without the if statement?

### DIFF
--- a/server.js
+++ b/server.js
@@ -2,19 +2,56 @@ import express from 'express'
 import bodyParser from 'body-parser'
 import cors from 'cors'
 import mongoose from 'mongoose'
+import topMusicData from './data/top-music.json'
 
-// If you're using one of our datasets, uncomment the appropriate import below
-// to get started!
-// 
 // import goldenGlobesData from './data/golden-globes.json'
 // import avocadoSalesData from './data/avocado-sales.json'
 // import booksData from './data/books.json'
 // import netflixData from './data/netflix-titles.json'
-// import topMusicData from './data/top-music.json'
+
 
 const mongoUrl = process.env.MONGO_URL || "mongodb://localhost/project-mongo"
 mongoose.connect(mongoUrl, { useNewUrlParser: true, useUnifiedTopology: true })
 mongoose.Promise = Promise
+
+const TopMusic = mongoose.model('TopMusic', {
+        id: {
+          type: Number},
+        trackName: {
+          type: String},
+        artistName: {
+          type: String},
+        genre: {
+          type: String},
+        bpm: {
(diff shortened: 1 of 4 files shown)


server.js
```diff
@@ -2,20 +2,51 @@ import express from 'express'
...
+        valence: {
+          type: Number},
+        length: {
+          type: Number},
+        acousticness: {
+          type: Number},
+        speechiness: {
+          type: Number},
+        popularity: {
+          typr: Number}
```
puj: Does your VSCode auto format like this?

server.js
```diff
@@ -29,7 +60,47 @@ app.use(bodyParser.json())
...
+//Using promises
+app.get('/topmusics', (req, res) => {
+  TopMusic.find()
+  .then((results) => {
+    // console.log('Found: ' + results)
+    res.json(results)
+  })
+})
+
+app.get('/topmusics/id/:id', (req, res) => {
```
puj: You could probably get away with `/topmusics/:id`

server.js
```diff
@@ -29,7 +60,47 @@ app.use(bodyParser.json())
...
+  const id = req.params.id
+  TopMusic.findOne({'id': id})
+  .then((results) => {
+    // console.log('Found: ' + results)
+    if (results === null)
+    throw "wrong" 
+    res.json(results)
+  }).catch((err) => {
+    console.log('Error' + err)
+    res.json({message: 'Cannot find this song', err: err})
```
puj: Consider changing the message to something like "Could not complete request" and let `err` explain more about why.

server.js
```diff
@@ -29,7 +60,47 @@ app.use(bodyParser.json())
...
+  })
+})
+
+// http://localhost:8080/topmusics/artist?q=ariana - How to search for artists
+
+app.get('/topmusics/artist', (req, res) => {
+  const queryString = req.query.q;
```
puj: Could there be a more descriptive name for your query parameter?

server.js
```diff
@@ -29,7 +60,47 @@ app.use(bodyParser.json())
...
 app.get('/', (req, res) => {
-  res.send('Hello world')
+  res.send('Music makes people come togheter!')
+})
+
+// Using Async + Await
+// app.get('/topmusics', async (req, res) => {
+//   const topmusic = await TopMusic.find()
+//   res.json(topmusic)
+//   })
```
puj: Did you like the promise approach more than async/await?

PR #12, opened 2020-02-10
Mongo API project - Anna Almestål

anna-ip:
So I think this week has been quit straight forward, nothing really that has been super hard to grasp.
I've made a couple of endpoints following Damien and Vans instructions and made four models. 
I choose to work with async and await and maybe that suddenly made sense.
I managed to get the pagination to work with skip and limit, so that was nice.

### DIFF
--- a/server.js
+++ b/server.js
@@ -28,7 +28,7 @@ const mongoUrl = process.env.MONGO_URL || "mongodb://localhost/project-mongo"
 mongoose.connect(mongoUrl, { useNewUrlParser: true, useUnifiedTopology: true })
 mongoose.Promise = Promise
 
-const Artist = mongoose.model('Artist', {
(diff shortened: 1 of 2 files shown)


anna-ip:
The deployed version:
[https://top-music-api.herokuapp.com/](url)

server.js
```diff
@@ -32,6 +87,74 @@ app.get('/', (req, res) => {
...
+})
+
+app.get('/records', async (req, res) => {
+  const record = await Record.find().populate('artist')
+  if (record) {
+    res.json(record)
+  } else {
+    res.status(404).json({ errror: 'Record not found' })
+  }
+})
```
puj: I'm having a hard time seeing what the functional difference between `/records` and `/artists` is

PR #14, opened 2020-02-10
Mongo Api - Linda

liissee:
I thought this week was hard and I had a hard time understanding the difference from last week. 

I tried to do as the codealong when you add a model with extra information to another model, but that didn't work when I used a json file.. I just got a lot of error messages! 
For example the House model, I want that information in the Character model. 
house: {
     type: mongoose.Schema.Types.ObjectId,
     ref: 'House'
   }

It that not possible? 

And I also found it hard to know when the error messages show.


README.md
```diff
@@ -1,13 +1,24 @@
 # Mongo API Project
 
-Replace this readme with your own information about your project. 
+In this project I have built and API using MongoDB and Mongoose models. The API has RESTful endpoints and returns arrays of data and a single character. 
 
-Start by briefly describing the assignment in a sentence or two. Keep it short and to the point.
+I have used Kaggle to find a dataset.
```
puj: Nice using a different dataset!

server.js
```diff
@@ -1,38 +1,167 @@
...
-// import booksData from './data/books.json'
-// import netflixData from './data/netflix-titles.json'
-// import topMusicData from './data/top-music.json'
-
-const mongoUrl = process.env.MONGO_URL || "mongodb://localhost/project-mongo"
+import mongoose, { mongo } from 'mongoose'
+import potterData from './data/potterData.json'
+
+//The name of the database - mongodb://localhost/mongo-project-potter
+const mongoUrl = process.env.MONGO_URL || "mongodb://localhost/mongo-project-potter"
```
puj: Good rename!

server.js
```diff
@@ -1,38 +1,167 @@
...
 app.use(bodyParser.json())
+//A function that recives a request, the response and the argument "next". 
+//It will execute before the routes below, if it's not evoked next() it will block the code coming next. The connection.readyState checks that the connection is good.
+app.use((req, res, next) => {
+  if (mongoose.connection.readyState === 1) {
+    next()
+  } else {
+    res.status(503).json({ error: 'Service unavailable' })
+  }
+})
```
puj: Interesting! This essentially checks if mongoose can establish a connection before trying to respond to any request. Hooray for custom middleware!

server.js
```diff
@@ -1,38 +1,167 @@
...
+  };
+  //Regular expression to make it case insensitive
+  if (req.query.name) { queryObj['name'] = new RegExp(req.query.name, 'i') }
+  if (req.query.house) { queryObj['house'] = new RegExp(req.query.house, 'i') }
+  if (req.query.job) { queryObj['job'] = new RegExp(req.query.job, 'i') }
+  if (req.query.gender) { queryObj['gender'] = new RegExp(req.query.gender, 'i') }
+
+  Character.find(queryObj)
+    .skip(pageOptions.page * pageOptions.limit)
+    .limit(pageOptions.limit)
```
puj: Your paging works really well, good job!

server.js
```diff
@@ -1,38 +1,167 @@
...
+    .sort({ id: 1 })
```
puj: Maybe there is another way to sort by default which would be more useful for the client?

server.js
```diff
@@ -1,38 +1,167 @@
...
+    .then((results) => {
+      // Succesfull
+      console.log('Found : ' + results);
+      res.json(results);
+    }).catch((err) => {
+      // Error/Failure
+      console.log('Error ' + err);
+      res.json({ message: 'Cannot find this character', err: err }); //Don't understand when this error message will show?
```
puj: This could happen if Mongoose can't complete the search for some reason.  

To handle not finding results, in your `.then()`

One option is to check if there are any results and return something different to the client.
Another option would be to check if there are no results and `throw` and error which would then execute whatever is in `.catch((err) => {...})`

server.js
```diff
@@ -1,38 +1,167 @@
...
+
+
+//GET SPECIFIC CHARACTER  http://localhost:9090/characters/2
+app.get("/characters/:id", async (req, res) => {
+  const id = req.params.id
+  const character = await Character.findOne({ "id": id })
+  if (character) {
+    res.json(character)
+  } else {
+    res.status(404).json({ error: "Character not found" })
```
puj: Nice status code usage

liissee:
Thank you! :) 


> > I tried to do as the codealong when you add a model with extra information to another model, but that didn't work when I used a json file.. I just got a lot of error messages!
> > For example the House model, I want that information in the Character model.
> > house: {
> > type: mongoose.Schema.Types.ObjectId,
> > ref: 'House'
> > }
> > It that not possible?
> 
> I'm sure it's possible but your .json needs to match the model you've created in Mongoose. Without seeing the errors etc, it's difficult to tell what went wrong.
> 
> I liked your approach and the variety of features you chose to include! Good job and keep up the good work 😄

Thank you! :) 
How do you mean it needs to match? 
I thought it was enough if you created a model with something that exists in the .json. Like "house" in this case..

//HOUSE MODEL
const House = mongoose.model('House', {
  name: String,
  mascot: String,
  head_teacher: String,
  ghost: String,
  founder: String
})

//POTTER CHARACTER MODEL
const Character = mongoose.model('Character', {
  id: Number,
  name: String,
  gender: String,
  job: String,
  **house: {
    type: mongoose.Schema.Types.ObjectId,
    ref: 'House'
  },** this seems to be the problem.. 
  wand: String,
  patronus: String,
  species: String,
  blood_status: String,
  hair_colour: String,
  eye_colour: String,
  loyalty: String,
  skills: String,
  birth: String,
  death: String
})

    const gryffindor = new House({ name: "Gryffindor", mascot: "Lion", head_teacher: "Minverva McGonogall", ghost: "Nearly Headless Nick", founder: "Godric Gryffindor" })
    await gryffindor.save()

//Route to see all characters with more house information
app.get('/', async (req, res) => {
  const characters = await Character.find().populate('house')
  res.json(characters)
})

this it the error message I get when RESET_DB=false npm run dev 
Skärmavbild 2020-02-12 kl  20 41 47

And this is a super long message when I have set RESET_DB=true npm run dev
Skärmavbild 2020-02-12 kl  20 54 11

Hope you understand what I mean 😄..

PR #15, opened 2020-02-10
Project Mongo API, Michel Elmoznino Laufer

MichelLaufer:
Deployed project:
https://michel-project-mongo.herokuapp.com/books
Frontend part:
https://michel-apiexpressproject.netlify.com/
https://github.com/MichelLaufer/project-api-express-books/tree/master/src

I think that the highlight of the week was the code-alongs. I think it is fun and very useful to learn backend. I had some technical difficulties to get started with this project getting relatively late into this project. I also had some initial difficulties to deploy the project. But these issues were resolved thanks to help from the teachers and some of the other students.

It would be nice to learn more about e.g. using .limit() and .skip() to allow the user to ask for the next page.


server.js
```diff
@@ -2,37 +2,113 @@ import express from 'express'
...
+  if(language) {
+   Book.find({'language_code': language})
+     .then((results) => {
+       console.log('Found')
+       res.json(results)
+      }) .catch((err) => {
+        console.log('Error ' + err)
+        res.json({message: 'Cannot find book', err: err}) 
+    })
+  }
```
puj: I tried `https://michel-project-mongo.herokuapp.com/books?language=none`
Did you expect your application to return no results or to give an error message?

server.js
```diff
@@ -2,37 +2,113 @@ import express from 'express'
...
+  Book.find({'title': queryRegex})
+    .sort({'average_rating': -1})
+    .then((results) => {
+      console.log('Found')
+      res.json(results)
+    }) .catch((err) => {
+       console.log('Error ' + err)
+       res.json({message: 'Cannot find book', err: err}) 
+    })
+})
```
puj: Usually it's a good idea to avoid too many/multiple exit points in a function.  Is there a way to restructure the function to only send responses in two cases (success/failure)?

server.js
```diff
@@ -2,37 +2,113 @@ import express from 'express'
...
+  const _id = req.params._id
+  Book.findOne({'_id': _id})
+    .then((results) => {
+      res.json(results)
+    }) .catch((err) => {
+      res.json({message: 'Cannot find this book', err: err})
+    })
+})
+
+
```
puj: Would be awesome to see an endpoint for top 30 books that you use in your frontend!

Creating an endpoint which gets a certain number of book and returns them sorted by rating using mongo could replace https://github.com/MichelLaufer/project-api-express-books/blob/master/src/App.js#L19

PR #17, opened 2020-02-10
mongo-api

horsefacegirl:
https://horsefacegirl-mongo-api.herokuapp.com/

### DIFF
--- a/server.js
+++ b/server.js
@@ -89,18 +89,6 @@ app.get("/directors/:id", async (req, res) => {
   res.json(director);
 });
 
-app.get("/directors/:id/titles", async (req, res) => {
-  const director = await Director.findById(req.params.id);
-  if (director) {
-    const titles = await Title.find({
-      director: mongoose.Types.ObjectId(director.id)
-    });
-    res.json(titles);
-  } else {
-    res.status(404).json({ error: "Director not found" });
-  }
-});
-
 app.get("/titles", async (req, res) => {
   const titles = await Title.find().populate("director");
   res.json(titles);
--- a/server.js
+++ b/server.js
@@ -3,24 +3,7 @@ import bodyParser from "body-parser";
 import cors from "cors";
 import mongoose from "mongoose";
 
-// import netflixData from "./data/netflix-titles.json";
-
-// Example json structure
-
-// {
-//   "show_id": 81193313,
-//   "title": "Chocolate",
-//   "director": "",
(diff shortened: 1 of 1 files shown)


server.js
```diff
@@ -1,38 +1,100 @@
...
+});
+
+app.get("/directors", async (req, res) => {
+  const directors = await Director.find();
+  res.json(directors);
+});
+
+app.get("/directors/:id", async (req, res) => {
+  const director = await Director.findById(req.params.id).populate("titles");
+  res.json(director);
```
puj: Well done. 

I would recommend one slight change in your REST API:

`/directors/:id` - return information about the director, without the associated titles
`/directors/:id/titles` - return the titles associated with this director (and optionally the info of the director)

server.js
```diff
@@ -1,38 +1,100 @@
...
+  type: String
+});
+
+const seedDatabase = async () => {
+  await Director.deleteMany();
+  await Title.deleteMany();
+
+  const distinct = (value, index, self) => {
+    return self.indexOf(value) === index;
+  };
```
puj: Extremely concise way of preventing duplicates.  Not readable if you've never seen it before though.  A comment is appropriate.

server.js
```diff
@@ -1,38 +1,100 @@
...
+
+  const uniqNetflixData = netflixData.filter(title => title.director !== "");
+
+  const directors = uniqNetflixData.map(title => title.director).filter(distinct);
```
puj: I like seeing the usage of filter and map

server.js
```diff
@@ -1,38 +1,100 @@
...
+
+  directors.forEach(async director => {
+    const newDirector = new Director({
+      _id: mongoose.Types.ObjectId(),
```
puj: Must this be explicitly specified?

server.js
```diff
@@ -1,38 +1,100 @@
...
+      name: director
+    });
+
+    const directorsTitles = netflixData.filter(
+      title => title.director !== "" && title.director === director
```
puj: Some parentheses could make this logic clearer/more readable

server.js
```diff
@@ -1,38 +1,100 @@
...
+
+    directorsTitles.forEach(async title => {
+      const newTitle = new Title({
+        title: title.title,
+        director: newDirector._id,
+        cast: title.cast,
+        country: title.country,
+        release_year: title.release_year,
+        type: title.type
+      });
```
puj: Is there duplicate data in the database where the `Title` information is the same but only the `director` field is different?

PR #21, opened 2020-02-10
project api

camillalof:
Hi,

I think I have to explain what my Api is, and its a Yoga pose API. The objects called Asanas are yoga-poses and the objects called Chakras is the field in the body that the poses affects. So I made a connection between the objects, and try to do it more or less like the pokemons and the different teams that Jennie showed us. 

I managed to see the information in postman all this time, and the ports worked. But I couldnt get the queries or the error messages to work even though I followed the code. Can I please get some feedback on this, It would be great to know what I did wrong. I heard that I could make a "throw"/"catch" for the error, but shouldnt the error just appear by writing the way I did it? :)

I also didnt see the information at the deployed site. Why can this be? I did put the RESET_DATABASE under config vars in heroku but is still is not working. Should I have written something more? 

Sorry that there where so many questions this week, i thought almost everything (except queries and errors where working until an hour ago).

Thanks for the help!


camillalof:
Here is the deplyed project (that is not working): https://api-project-yoga-poses.herokuapp.com/

server.js
```diff
@@ -3,19 +3,120 @@ import bodyParser from 'body-parser'
...
+  number: Number,
+  color: String
+})
+
+const Asana = mongoose.model('Asana', {
+  name: String,
+  advanced: Boolean,
+  chakra: {
+  type: mongoose.Schema.Types.ObjectId,
+  ref: 'Chakra'
```
JennieDalgren: indentation :)

roy2git:
I have developed a simple MERN stack project for yoga poses. Pls check if that is of any use to you https://yogaposes.herokuapp.com/

PR #22, opened 2020-02-10
PR - Tiago Ferreira

tavferreira:
**Link to my deployed project:**
https://tavferreira-mongo-api.herokuapp.com/

**My reflections on how this week's project turned out:**
The concept of seeding the data into the database and retrieve it with the db methods was quite easy to absorb, but in practice it was hard to implement it correctly. 
My idea was to have categories and films as models also, but I couldn't find a way to seed the data correctly so the ids were linked correctly.
In the end, I'm happy that at least I managed to do it for nominees.

**Things I'd like to have clarified or explained in more detail:** 
How to seed data with more models and avoid duplications.


server.js
```diff
@@ -2,20 +2,62 @@ import express from 'express'
...
+})
+
+const Nomination = mongoose.model('Nomination', {
+  year_film: Number,
+  year_award: Number,
+  ceremony: Number,
+  category: String,
+  nominee: {
+    type: mongoose.Schema.Types.ObjectId,
+    ref: 'Nominee'
```
JennieDalgren: nice ⭐️

PR #23, opened 2020-02-11
Matilda Arvidsson project mongoDB

HIPPIEKICK:
Sorry for submitting late. I'm very happy I managed to use the mongo-methods for the queries. The documentation was so confusing with very few examples haha. Can you confirm that I understood correctly in the readme, regarding "The problem"? Like, if you use mongo-methods you kind of sort before fetching everything? It appeared to me that every request was faster after switching to mongo-methods instead of first using find() to get the whole Model and then sorting with vanilla JS-methods.

Last week I had something like this - if someone writes ?page=3 and there are only 2 pages, then the second page should show.
```
  if (pageSearch > pageCount) {
    pageSearch = pageCount
  }
```
how could I have done that this week?

Also, I found another dataset with more info but the same id's. Would it be possible to connect these two datasets in some way or more logical - populate the dbModel with two different datasets? *dreaming*

Thanks in advance!


server.js
```diff
@@ -3,33 +3,119 @@ import bodyParser from 'body-parser'
...
 app.use(cors())
 app.use(bodyParser.json())
 
-// Start defining your routes here
+//Database not connected
+app.use((req, res, next) => {
+  if (mongoose.connection.readyState === 1) {
+    next()
+  } else {
+    res.status(503).json({ error: "Service unavailable" })
```

PR #27, opened 2020-02-12
mongo-api-project

Agnes94:
https://mongo-api-project.herokuapp.com/


server.js
```diff
@@ -20,19 +20,104 @@ mongoose.Promise = Promise
 // overridden when starting the server. For example:
 //
 //   PORT=9000 npm start
```
JennieDalgren: you can remove all unused code

PR #30, opened 2020-02-13
books-mongo-api

JoacimNilsen:
This week was hard! Mostly getting to set up the mongoDB wich I am still trying to fix. My deployed site does not run yet. I managed to set it up running on local server so that I could move on with my project. I will try to fix this on my own but I might need some help . 
I also tried making an endpoint that could sort out books by starting letter of titles with regEx but did not succeed. Could not put more time on that so I skipped it for now!

Here is a link to my deployed project:
https://joacim-mongo-books.herokuapp.com/

### DIFF
--- a/server.js
+++ b/server.js
@@ -40,14 +40,14 @@ if(process.env.RESET_DB) {
   seedDatabase()
 }
 
-//Return Error if connection fail
-  app.use((req, res, next) => {
-  if (mongoose.connection.readyState === 1) {
-    next()
-  } else {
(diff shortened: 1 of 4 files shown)


server.js
```diff
@@ -2,37 +2,95 @@ import express from 'express'
...
+}
+
+//Return Error if connection fail
+  app.use((req, res, next) => {
+  if (mongoose.connection.readyState === 1) {
+    next()
+  } else {
+    res.status(503).json({ error: 'Service unavailable' })
+  }
+})
```

PR #31, opened 2020-02-13
Mongo DB API J Fick

Jenfi:
Having troubles with a lot, so now handing it in after trying by myself, in the hopes of getting some helpful feedback :)  

- Mongo atlas is not fetching/finding any collections although I have followed the guide carefully.
- Therefor (I think)  Heroku is not working either
- In my code I am trying to fetch errors, console an error message and updatee status in different ways of writing the code but they are not working.

### DIFF
--- a/server.js
+++ b/server.js
@@ -49,7 +49,7 @@ if (process.env.RESET_DB) {
 
     booksData.forEach((book) => {
       new Book(book).save()
-    });
+    })
   }
   seedDatabase()
 }
@@ -88,8 +88,6 @@ app.get('/books', (req, res) => {
     }
   })
 })
-
-
 
 // Book.find({ 'title': queryRegex })
 //   .then((results) => {
--- a/server.js
+++ b/server.js
@@ -2,17 +2,8 @@ import express from 'express'
 import bodyParser from 'body-parser'
 import cors from 'cors'
 import mongoose from 'mongoose'
-
-// If you're using one of our datasets, uncomment the appropriate import below
-// to get started!
-// 
-// import goldenGlobesData from './data/golden-globes.json'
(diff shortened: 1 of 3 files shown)


server.js
```diff
@@ -2,37 +2,128 @@ import express from 'express'
...
+    type: String
+  },
+  authors: {
+    type: String
+  },
+  average_rating: {
+    type: Number
+  },
+  isbn: {
+    // unique: true,
```
JennieDalgren: why not unique: true?

server.js
```diff
@@ -2,37 +2,128 @@ import express from 'express'
...
+  },
+  ratings_count: {
+    type: Number
+  },
+  text_reviews_count: {
+    type: Number
+  },
+})
+
+if (process.env.RESET_DB) {
```
JennieDalgren: Do be able to seed your deployed database you need to add this to heroku. could be the issue with not getting any data in atlas?

server.js
```diff
@@ -2,37 +2,128 @@ import express from 'express'
...
 app.use(bodyParser.json())
 
-// Start defining your routes here
+app.use((req, res, next) => {
+  if (mongoose.connection.readyState === 1) {
+    next()
+  } else {
+    res.status(503).json({ error: 'Service unavailble' })
+  }
+})
```

server.js
```diff
@@ -2,37 +2,128 @@ import express from 'express'
...
+
+// Book.find({ 'title': queryRegex })
+//   .then((results) => {
+//     console.log('Found ' + results);
+//     res.json(results);
+//   }).catch((err) => {
+//     console.log('Error ' + err);
+//     res.json({ message: 'Cannot find this book' })
+//   })
```
JennieDalgren: clean this up

PR #32, opened 2020-02-14
swolfson project mongo api

swolfson1:
https://project-mongo-swolfson.herokuapp.com/

### DIFF
--- a/server.js
+++ b/server.js
@@ -62,37 +62,39 @@ const app = express();
 app.use(cors());
 app.use(bodyParser.json());
 
-// Start defining your routes here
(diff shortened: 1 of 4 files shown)


server.js
```diff
@@ -1,38 +1,100 @@
...
+    type: String
+  }
+});
+
+const addMovieToDatabase = () => {
+  netflixData.forEach(movie => {
+    new Movie(movie).save();
+  });
+};
+addMovieToDatabase();
```
JennieDalgren: careful. this will add to the database every time the server starts. And will give you duplicates.

PR #33, opened 2020-02-18
Late PR after week 18 - Mongo DB - Jola Chabik

ucha3i:


server.js
```diff
@@ -3,33 +3,62 @@ import bodyParser from 'body-parser'
...
+  await Movie.deleteMany()
+
+  const cat = new Category({ name: 'Best movie of XX century' })
+  await cat.save()
+
+  const mov = new Movie({ title: 'Rejs (English title - The Cruise)', category: cat })
+  await mov.save()
+}
+
+seedDatabase()
```
JennieDalgren: careful though, as this will reset and seed the database every time you start the server

PR #35, opened 2020-03-04
Proj mongo-api

visua77:
Hade kanske glömt en pull här månne?

### DIFF
--- a/server.js
+++ b/server.js
@@ -5,32 +5,7 @@ import mongoose from 'mongoose'
 
 import netflixData from './data/netflix-titles.json'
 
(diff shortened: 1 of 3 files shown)


server.js
```diff
@@ -3,19 +3,37 @@ import bodyParser from 'body-parser'
...
+
+const seedDatabase = async() => {
+    await Show.deleteMany()
+    netflixData.forEach(show => {
+
+        new Show(show).save()
+    })
+}
+
+seedDatabase()
```
JennieDalgren: careful though, as this will reset and seed the database every time you start the server

PR #38, opened 2020-05-13
Project_18-Mongo_API

palhamel:
Link to my deployed project:
https://project-w18-mongo-api.herokuapp.com/

Interesting week, a mix a fun work and complicated things. For example when building the model and seeding database, you need to be careful or things can get messy. If I had more time I would like to do more sorting/filtering, routes and use my own data. 

Links to endpoints are in the README


server.js
```diff
@@ -1,38 +1,155 @@
...
+  num_pages: {
+    type: Number,
+  },
+  ratings_count: {
+    type: Number,
+  },
+  text_reviews_count: {
+    type: Number,
+  },
+});
```
KarolCislowski, line 48: Model created this way allows the user to add empty records to your database. 
Use `required: true` in some fields like title and authors to make it safe.

And one suggestion: mongoose generates unique _ID field for you. Do you really need to add bookID from the data file? I think it will be difficult to increment when you will need to add new records in the future :)

server.js
```diff
@@ -1,38 +1,155 @@
...
+  res.send(
+    "My API endpoints: /books ,/books/439554934 ,/sort?sort_by=average_rating ,/info/5eba83e3b3b1c00023249494  plus handling errors"
+  );
+});
+
+// To return all Books:
+// http://localhost:8080/books
+app.get("/books", async (req, res) => {
+  const books = await Book.find();
+  console.log(`Message: Found ${books.length} books`);
```
KarolCislowski, line 96: I think this line was created for testing, so It is no need to leave it in the final project :)

KarolCislowski, line 96: I think this line was created for testing, but it is no need to left it in the final project :)

PR #40, opened 2020-05-13
Project Mongo Api Hanna R 

Ruterryter:
It was a fun project, but i was sick this week so I did not spend so much time on it as I had wanted. 
https://hitsongsinmay.herokuapp.com/

### DIFF
--- a/server.js
+++ b/server.js
@@ -2,25 +2,41 @@ import express from 'express'
 import bodyParser from 'body-parser'
 import cors from 'cors'
 import mongoose from 'mongoose'
+import topMusicData from './data/top-music.json'
 
-// If you're using one of our datasets, uncomment the appropriate import below
-// to get started!
-// 
-// import goldenGlobesData from './data/golden-globes.json'
-// import avocadoSalesData from './data/avocado-sales.json'
-// import booksData from './data/books.json'
-// import netflixData from './data/netflix-titles.json'
-// import topMusicData from './data/top-music.json'
 
-const mongoUrl = process.env.MONGO_URL || "mongodb://localhost/project-mongo"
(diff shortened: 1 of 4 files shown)


server.js
```diff
@@ -2,37 +2,85 @@ import express from 'express'
...
 
-// Defines the port the app will run on. Defaults to 8080, but can be 
-// overridden when starting the server. For example:
-//
-//   PORT=9000 npm start
-const port = process.env.PORT || 8080
+
+const Song = mongoose.model('Song', {
+  id: {
+    type: Number
```
KarolCislowski, line 15: Mongoose adds unique _id field so it is no need to add one by hand. This will make your work harder when you will start adding new records to the database later.

server.js
```diff
@@ -2,37 +2,85 @@ import express from 'express'
...
+  },
+  trackName: {
+    type: String
+  },
+  artistName: {
+    type: String
+  },
+  genre: {
+    type: String
+  }
```
KarolCislowski, line 25: It will be safer for you if you add `required: true` to some fields. this will prevent from creating empty entries in the database ;)

server.js
Ruterryter: Ok, I thought we were supposed to not use that id since that will change when reseting the database, that this way is safer?

server.js
Ruterryter: Great, I will add that. :)

server.js
KarolCislowski: In the real-life project, you will not type the id by hand to extract the data. You will use app where you will take a full list of songs, choose a song by title, and under the hood, it will check the _id field and will ask API for details for this _id :)

And you will not reset the database after you make the project public :)

Ruterryter: Ok, great, then I think I get it :). Thanks for the explanation.

PR #41, opened 2020-05-13
review wanted

kajsabootin:
Link to my deployed project:

https://mongo-api-kajsa.herokuapp.com/

The thing I was most struggling about is when i should run "RESET_DB=true npm start" in the terminal.


server.js
```diff
@@ -1,38 +1,71 @@
...
+mongoose.connect(mongoUrl, { useNewUrlParser: true, useUnifiedTopology: true });
+mongoose.Promise = Promise;
+
+const Track = mongoose.model("Track", {
+  artistName: String,
+  trackName: String,
+  genre: String,
+  popularity: Number,
+  length: Number,
+  danceability: Number,
```
KarolCislowski, line 22: It will be safer for you if you add required: true to some fields. this will prevent from creating empty entries in the database ;)

server.js
```diff
@@ -1,38 +1,71 @@
...
+  id: Number,
```
KarolCislowski, line 23: Mongoose adds unique _id field so it is no need to add one by hand. This will make your work harder when you will start adding new records to the database later.

PR #43, opened 2020-05-13
Sara (review wanted)

Aqveduktis:
Live:
https://aqveduktis-mongo-api.herokuapp.com
deploying the database was a bit tricky


models.js
```diff
@@ -0,0 +1,13 @@
...
+
+export const ArtistDetail = mongoose.model('ArtisDetail', {
+ id: Number,
+ name: String,
+ years: String,
+ genre: String,
+ nationality: String,
+ bio: String,
+ wikipedia: String,
+ paintings: Number
```
KarolCislowski, line 12: It will be safer for you if you add required: true to some fields. this will prevent from creating empty entries in the database ;)

models.js
```diff
@@ -0,0 +1,13 @@
+import mongoose from 'mongoose'
+
+
+export const ArtistDetail = mongoose.model('ArtisDetail', {
+ id: Number,
```
KarolCislowski, line 5: Mongoose adds unique _id field so it is no need to add one by hand. This will make your work harder when you will start adding new records to the database later.

server.js
```diff
@@ -2,20 +2,26 @@ import express from 'express'
 import bodyParser from 'body-parser'
 import cors from 'cors'
 import mongoose from 'mongoose'
+import artistData from './data/artistsCSV.json'
+import {ArtistDetail} from './models'
```
KarolCislowski, line 6: Nice that you extracted models to separate file 👍

server.js
```diff
@@ -26,10 +32,59 @@ const app = express()
...
+    }
+    res.status(404).json({error:"artist not found"})
+  }
+  catch{
+    res.status(400).json({error:"invalid user ID"})
+  }
+})
+app.get('/artists/name/:name', async (req, res)=>{
+  const {name} = req.params
+  const queryRegex = new RegExp(name, 'i');
```
KarolCislowski, line 82: RegExp! 👍

PR #44, opened 2020-05-13
MongoDB – Kajsa Holmström (review wanted)

crushoftheyear:
This week I struggled with wrapping my head around the mongo methods, and not being able to do it like we did last week with the Express API project. I'm pleased with the result (because I feel that I really grasp it) but I would have liked to make the pages stretch goal. I created a couple of query parameters and error handling though.

Please view the Readme for documentation of endpoints/queries:
https://github.com/crushoftheyear/project-mongo-api

**Deployed**
https://mongo-api-netflix.herokuapp.com/


server.js
```diff
@@ -2,36 +2,100 @@ import express from 'express'
...
-
-const mongoUrl = process.env.MONGO_URL || "mongodb://localhost/project-mongo"
+const mongoUrl = process.env.MONGO_URL || 'mongodb://localhost/netflix'
 mongoose.connect(mongoUrl, { useNewUrlParser: true, useUnifiedTopology: true })
 mongoose.Promise = Promise
 
+const Show = mongoose.model('Show', {
+  show_id: {
+    type: Number
+  },
```

server.js
```diff
@@ -2,36 +2,100 @@ import express from 'express'
...
+  duration: {
+    type: String
+  },
+  description: {
+    type: String
+  },
+  type: {
+    type: String
+  }
```
KarolCislowski, line 29: It will be safer for you if you add required: true to some fields. this will prevent from creating empty entries in the database ;)

server.js
```diff
@@ -2,36 +2,100 @@ import express from 'express'
...
+  res.send(listEndpoints(app))
+})
+
+// Get all shows
+app.get('/shows', async (req, res) => {
+  const { title, type } = req.query
+
+  // Regex to make query case insensitive
+  const regex = (query) => {
+    return new RegExp(query, 'i')
```
KarolCislowski, line 69: RegExp! 👍

PR #45, opened 2020-05-13
project mongo api Anna Takman

annatakman:
The hardest thing this week was all the new programs that we had to download and install, it was quite unclear in the beginning what MongoDB was and how to use it. Writing the code was easier than all the installations. :)

### DIFF
--- a/server.js
+++ b/server.js
@@ -2,19 +2,64 @@ import express from 'express'
 import bodyParser from 'body-parser'
 import cors from 'cors'
 import mongoose from 'mongoose'
-
-// If you're using one of our datasets, uncomment the appropriate import below
-// to get started!
-// 
-// import goldenGlobesData from './data/golden-globes.json'
-// import avocadoSalesData from './data/avocado-sales.json'
-// import booksData from './data/books.json'
-// import netflixData from './data/netflix-titles.json'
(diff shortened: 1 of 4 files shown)


server.js
```diff
@@ -2,20 +2,62 @@ import express from 'express'
...
 
 const mongoUrl = process.env.MONGO_URL || "mongodb://localhost/project-mongo"
 mongoose.connect(mongoUrl, { useNewUrlParser: true, useUnifiedTopology: true })
 mongoose.Promise = Promise
 
+//Mongoose model 
+const Show = mongoose.model("Show", {
+  show_id: {
+    type: Number
+  },
```
KarolCislowski, line 15: Mongoose adds unique _id field so it is no need to add one by hand. This will make your work harder when you will start adding new records to the database later.

server.js
```diff
@@ -2,20 +2,62 @@ import express from 'express'
...
+  listed_in: {
+    type: String
+  },
+  description: {
+    type: String
+  },
+  type: {
+    type: String
+  }
```
KarolCislowski, line 45: It will be safer for you if you add required: true to some fields. this will prevent from creating empty entries in the database ;)

PR #46, opened 2020-05-13
Mongo API - Review wanted - Frida

fridamaria:
https://fridamaria-mongo-api.herokuapp.com/

This week was harder to grasp compared to last week. Still fun though. I tried creating almost the same endpoints as last week to ensure I understand how to get the same result in different ways. Switched up id for ISBN13 as it allowed me to try out some different validation aspects. I like how it turned out :) No frontend yet though, that would probably be the next step!

Looking forward to learning how to use post now!

### DIFF
--- a/models/book.js
+++ b/models/book.js
@@ -4,31 +4,33 @@ const Schema = mongoose.Schema
 
 const BookSchema = new Schema({
   bookID: {
-    type: Number,
+    type: Number
   },
   title: {
-    type: String,
+    type: String
   },
   authors: {
-    type: String,
+    type: String
   },
   average_rating: {
-    type: Number,
+    type: Number
   },
   isbn13: {
     type: Number,
+    min: 9780000000000,
+    max: 9999999999999
   },
   language_code: {
-    type: String,
+    type: String
   },
   num_pages: {
-    type: Number,
+    type: Number
   },
   ratings_count: {
-    type: Number,
+    type: Number
   },
   text_reviews_count: {
-    type: Number,
+    type: Number
   },
 })
 
--- a/models/book.js
+++ b/models/book.js
@@ -3,14 +3,13 @@ import mongoose from 'mongoose'
 const Schema = mongoose.Schema
 
 const BookSchema = new Schema({
-  bookID: {
-    type: Number
-  },
   title: {
-    type: String
+    type: String,
+    required: true,
   },
   authors: {
-    type: String
+    type: String,
+    required: true
   },
   average_rating: {
     type: Number
@@ -18,7 +17,8 @@ const BookSchema = new Schema({
   isbn13: {
     type: Number,
     min: 9780000000000,
-    max: 9999999999999
+    max: 9999999999999,
+    required: true
   },
   language_code: {
     type: String
--- a/models/book.js
+++ b/models/book.js
@@ -31,7 +31,7 @@ const BookSchema = new Schema({
   },
   text_reviews_count: {
     type: Number
-  },
+  }
 })
 
 module.exports = new mongoose.model('Book', BookSchema)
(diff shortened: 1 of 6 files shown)


models/book.js
```diff
@@ -0,0 +1,37 @@
+import mongoose from 'mongoose'
+
+const Schema = mongoose.Schema
+
+const BookSchema = new Schema({
+  bookID: {
+    type: Number
+  },
```
KarolCislowski, line 8: Mongoose adds unique _id field so it is no need to add one by hand. This will make your work harder when you will start adding new records to the database later.

models/book.js
```diff
@@ -0,0 +1,37 @@
...
+  num_pages: {
+    type: Number
+  },
+  ratings_count: {
+    type: Number
+  },
+  text_reviews_count: {
+    type: Number
+  },
```
KarolCislowski, line 34: It will be safer for you if you add required: true to some fields. this will prevent from creating empty entries in the database ;)

Good that you added range for ISBN number :) 👍 

text_reviews_count is your last field, so you shouldn't add coma after it in line 34

server.js
```diff
@@ -2,37 +2,128 @@ import express from 'express'
...
-
-// If you're using one of our datasets, uncomment the appropriate import below
-// to get started!
-// 
-// import goldenGlobesData from './data/golden-globes.json'
-// import avocadoSalesData from './data/avocado-sales.json'
-// import booksData from './data/books.json'
-// import netflixData from './data/netflix-titles.json'
-// import topMusicData from './data/top-music.json'
+import Book from './models/book'
```
KarolCislowski, line 5: Good job! The external model file makes your project cleaner and easier to read! 👍

server.js
```diff
@@ -2,37 +2,128 @@ import express from 'express'
...
+    } else if (sort === 'rating_asc') {
+      return { average_rating: 1 }
+    } else if (sort === 'pages_dsc') {
+      return { num_pages: -1 }
+    } else if (sort === 'pages_asc') {
+      return { num_pages: 1 }
+    } else if (sort === 'alphabetical') {
+      return { title: 1 }
+    }
+  }
```
KarolCislowski, line 69: Nice sorting variety!

server.js
```diff
@@ -2,37 +2,128 @@ import express from 'express'
...
+    authors: new RegExp(author, 'i')
+  })
+    .sort(sortQuery(sort))
+    .limit(perPage)
+    .skip(skip)
+
+  // Used to display total books, checking booksList length without pagination
+  const booksListNoPagination = await Book.find({
+    title: new RegExp(title, 'i'),
+    authors: new RegExp(author, 'i')
```
KarolCislowski, line 84: RegExp 👍

server.js
```diff
@@ -2,37 +2,128 @@ import express from 'express'
...
+  }
+})
+
+// Route for single book using ISBN-13 as param
+app.get('/books/:isbn13', async (req, res) => {
+  const { isbn13 } = req.params
+  const error404 = `No book found with ISBN-13 ${isbn13}`
+  const error400 = `${isbn13} is not a valid ISBN-13`
+  // Used to check if the isbn13 param starts with 978 followed by 10 digits
+  const isbnCheck = /^(978)([0-9]{10})$/
```
KarolCislowski, line 107: Nice usage of regular expression!

models/book.js
fridamaria: Ah, makes sense! Added it to some of them :)

Removed the comma :)

models/book.js
fridamaria: Removed it :)

server.js
fridamaria: I was happy when it worked out, so thanks!

And thank you for the fastest review in history 🎉

KarolCislowski: You're welcome :)

PR #47, opened 2020-05-13
Project 18 - Johanna Åstrand - Review wanted

jastrand:
This project has been fun - I feel that I could have given it a bit more effort but it has been a busy week for me.

### DIFF
--- a/server.js
+++ b/server.js
@@ -68,9 +68,23 @@ app.get('/', (req, res) => {
 })
 
 app.get('/books', async (req, res) => {
-  const books = await Book.find()
+  const { query } = req.query;
+  const queryRegex = new RegExp(query, 'i')
+  const books = await Book.find({ title: queryRegex }).sort({
+    average_rating: -1,
+  })
   res.json(books)
 })
+
+app.get('/books/:isbn', async (req, res) => {
+  const { isbn } = req.params;
+  const book = await Book.findOne({ isbn: isbn });
+  if (book) {
+    res.json(book);
+  } else {
+    res.status(404).json({ error: `Could not find book with isbn=${isbn}` });
+  }
+});
 
 // Start the server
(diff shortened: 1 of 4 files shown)


server.js
```diff
@@ -2,36 +2,103 @@ import express from 'express'
...
-
-const mongoUrl = process.env.MONGO_URL || "mongodb://localhost/project-mongo"
+const mongoUrl = process.env.MONGO_URL || "mongodb://localhost/books"
 mongoose.connect(mongoUrl, { useNewUrlParser: true, useUnifiedTopology: true })
 mongoose.Promise = Promise
 
+const Book = mongoose.model('Book', {
+  bookID: {
+    type: Number,
+  },
```

server.js
```diff
@@ -2,36 +2,103 @@ import express from 'express'
...
+  num_pages: {
+    type: Number,
+  },
+  ratings_count: {
+    type: Number,
+  },
+  text_reviews_count: {
+    type: Number,
+  },
```
KarolCislowski, line 41: It will be safer for you if you add required: true to some fields. this will prevent from creating empty entries in the database ;)

text_reviews_count is the last field of the model so you shouldn't add coma after it (line 41)

server.js
```diff
@@ -2,36 +2,103 @@ import express from 'express'
...
+
 // Start defining your routes here
 app.get('/', (req, res) => {
-  res.send('Hello world')
+  res.send('Welcome to my BOOKS database! Check the readme for endpoint documentation')
+})
+
+app.get('/books', async (req, res) => {
+  const { query } = req.query;
+  const queryRegex = new RegExp(query, 'i')
```
KarolCislowski, line 81: RegExp 👍

server.js
jastrand: Great input!

PR #50, opened 2020-05-14
project-mongo Anne-Sophie (review wanted)

annesophiegendron:
https://annesophie-mongo-project.herokuapp.com/lists

I had some issues to deploy on Heroku. Just found out that I forgot to add  RESET_DATABASE as a key and value to true to make it work. I am so happy it finally works

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,13 +1,18 @@
 # Mongo API Project
+Build a Mongo database
 
-Replace this readme with your own information about your project. 
+## Learnings
+- Create routes
+- Create mongoose models
+- Deploy Node projects
 
-Start by briefly describing the assignment in a sentence or two. Keep it short and to the point.
-
-## The problem
-
-Describe how you approached to problem, and what tools and techniques you used to solve it. How did you plan? What technologies did you use? If you had more time, what would be next?
+## TECH
+- Heruko
+- Express
+- Mongo DB
+- Mongoose
+- Postman
+- Javascript
 
 ## View it live
-
-Every project should be deployed somewhere. Be sure to include the link to the deployed project so that the viewer can click around and see what it's all about.
+https://netflix-mongo-annesophie.herokuapp.com/
--- a/README.md
+++ b/README.md
@@ -1,5 +1,8 @@
 # Mongo API Project
-Build a Mongo database
+Build a Mongo database.
+Possible routes: /shows /lists /directors
+Combinable query by type: /shows?type=movie, /directors/:id
+
 
 ## Learnings
 - Create routes
--- a/README.md
+++ b/README.md
@@ -15,4 +15,4 @@ Build a Mongo database
 - Javascript
 
 ## View it live
-https://netflix-mongo-annesophie.herokuapp.com/
+https://annesophie-mongo-project.herokuapp.com/
--- a/server.js
+++ b/server.js
@@ -9,8 +9,9 @@ mongoose.connect(mongoUrl, { useNewUrlParser: true, useUnifiedTopology: true })
 mongoose.Promise = Promise
 
 const Show = mongoose.model('Show', {
-  title: { type: String },
-  director: { type: String },
+  show_id: { type: Number },
+  title: { type: String, required: true },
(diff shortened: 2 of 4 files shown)


README.md
```diff
@@ -1,13 +1,18 @@
...
-## The problem
-
-Describe how you approached to problem, and what tools and techniques you used to solve it. How did you plan? What technologies did you use? If you had more time, what would be next?
+## TECH
+- Heruko
+- Express
+- Mongo DB
+- Mongoose
+- Postman
+- Javascript
```
KarolCislowski, line 16: Your documentation does not show how to use your API. 
Adding the list of endpoints here would be a great enchantment :)

server.js
```diff
@@ -2,36 +2,125 @@ import express from 'express'
...
+  director: { type: String },
+  cast: { type: String },
+  country: { type: String },
+  date_added: { type: String },
+  release_year: { type: Number },
+  rating: { type: String },
+  duration: { type: String },
+  listed_in: { type: String },
+  description: { type: String },
+  type: { type: String }
```
KarolCislowski, line 22: It will be safer for you if you add required: true to some fields. this will prevent from creating empty entries in the database ;)

server.js
```diff
@@ -2,36 +2,125 @@ import express from 'express'
...
+})
+
+const Title = mongoose.model('Title', {
+  title: String,
+  _id: Number,
+  director: {
+    type: mongoose.Schema.Types.String,
+    ref: 'Director'
+  }
+})
```
KarolCislowski, line 40: Great Job! References are crucial to building larger systems 👍

server.js
```diff
@@ -2,36 +2,125 @@ import express from 'express'
...
 
-// Start defining your routes here
+// if database not connected
+app.use((req, res, next) => {
+  if (mongoose.connection.readyState === 1) {
+    next()
+  } else {
+    res.status(503).json({ error: 'Service unavailable' })
+  }
+})
```
KarolCislowski, line 72: Great database connection test!

server.js
```diff
@@ -2,36 +2,125 @@ import express from 'express'
...
+  director: String,
+  title: {
+    type: mongoose.Schema.Types.String,
+    ref: 'Title'
+  }
+})
+
+const Title = mongoose.model('Title', {
+  title: String,
+  _id: Number,
```
KarolCislowski, line 35: Mongoose adds unique _id field so it is no need to add one by hand. This will make your work harder when you will start adding new records to the database later.

server.js
```diff
@@ -2,36 +2,125 @@ import express from 'express'
...
+})
+
+app.get('/titles', async (req, res) => {
+  const titles = await Title.find().populate('director')
+  if (titles) {
+    res.json(titles)
+  } else {
+    res.status(404).json({ error: 'Title show not found' })
+  }
+})
```
KarolCislowski, line 122: This route doesn't work. Probably because you added the _id field manually, try to remove it from the schema and reseed your database.

README.md
annesophiegendron: updated :)

server.js
annesophiegendron: It is still returning an empty array and I don't get why because it is the exact same mongoose model as I did for 'Director'. I removed the manual id I had before and set "required=true" in the 'Show' model

KarolCislowski: I found why this endpoint is not working. If you add something after the  `find()` like `populate()` you need to use `exec()` on the end of the chain :)

server.js
```diff
@@ -2,34 +2,144 @@ import express from 'express'
...
+app.get('/directors/:id', async (req, res) => {
+  const director = await Director.findById(req.params.id)
+  if (director) {
+    res.json(director)
+  } else {
+    res.status(404).json({ error: "Director not found" })
+  }
+})
+
+app.get('/title', async (req, res) => {
```
KarolCislowski, line 127: Change `title` to `titles`, please, because REST endpoints are written in plural form and everything will be great!

PR #51, opened 2020-05-14
Camilla project mongo api

camillagaus:


server.js
```diff
@@ -2,37 +2,61 @@ import express from 'express'
...
 const app = express()
 
-// Add middlewares to enable cors and json body parsing
 app.use(cors())
 app.use(bodyParser.json())
 
-// Start defining your routes here
+
+const Artist = mongoose.model('Artist', {
+  id: String,
```
KarolCislowski, line 19: Mongoose adds unique _id field so it is no need to add one by hand. This will make your work harder when you will start adding new records to the database later.

server.js
```diff
@@ -2,37 +2,61 @@ import express from 'express'
...
+  trackName: String,
+  artistName: String,
+  genre: String,
+  danceability: Number
```
KarolCislowski, line 23: It will be safer for you if you add `required: true` to some fields. this will prevent from creating empty entries in the database ;)

server.js
```diff
@@ -2,37 +2,61 @@ import express from 'express'
...
+  console.log('resetting the database')
+const seedDatabase = async () => {
+  await Artist.deleteMany()
+  
+  topMusicData.forEach((artist) => {
+    new Artist(artist).save()
+  })
+}
+seedDatabase()
+//}
```
KarolCislowski, line 36: Why you commented out your "if statement"? This function needs to be invoked conditionally.

PR #52, opened 2020-05-14
Mongo API - Josephine

prosperpine:
I thought it was really difficult this week. There were a lot of new tools that came with Mongo and my head felt like a mess at times. Regarding the endpoints, I struggled with returning a single result since I chose the Golden Globes data and the objects doesn't have an id, and I didn't want to use only the mongodb-object-id. So I tried to find something that would capture the winner of a specific category in a specific year, since there is only one winner. I know the code is probably not best practice, I tried to use the RegExp for this endpoint as well but couldn't make it work during the time I had to spend on this project. So this is probably what I would have wanted to have done differently if I had more time. 

https://first-mongodb-project.herokuapp.com/

### DIFF
--- a/server.js
+++ b/server.js
@@ -12,9 +12,11 @@ const Nominee = mongoose.model('Nominee', {
   nominee: {
     type: String,
   },
+  ceremony: {
+    type: Number,
+  },
   category: {
-    type: mongoose.Schema.Types.String,
-    ref: 'Category',
+    type: String,
   },
   year_award: {
     type: Number,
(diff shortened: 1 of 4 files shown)


prosperpine:
Hello Karol,

I added one route for a single nomination now, thet you can get from id.
Will you look at it again?

Thanks,

Josephine

Den mån 18 maj 2020 kl 15:01 skrev Karol Cislowski <notifications@github.com

server.js
```diff
@@ -1,38 +1,114 @@
...
 
 // Start defining your routes here
 app.get('/', (req, res) => {
-  res.send('Hello world')
-})
+  res.send('Hello world');
+});
+
+// Shows a list of all the winners a certain year
+app.get('/year/:year/winners', async (req, res) => {
```
KarolCislowski, line 54: change year to years- in REST apis we use plural forms.

server.js
```diff
@@ -1,38 +1,114 @@
...
+    res.json(nomination);
+  } else {
+    res.status(404).json({ error: 'nomination not found' });
+  }
+});
+
+// Shows the winner of a certain category from a specific ceremony
+// example:/winner/ceremonies/68/categories/best motion picture - drama
+app.get(
+  '/winner/ceremonies/:ceremony/categories/:category',
```
KarolCislowski, line 78: Here also should be 'winners'

PR #53, opened 2020-05-14
Mongo API Emma

emmapre:
I would like some clarification why the regex doesn't work. Also why the findById doesn't work.


server.js
```diff
@@ -3,35 +3,122 @@ import bodyParser from 'body-parser'
...
-// import topMusicData from './data/top-music.json'
+import netflixData from './data/netflix-titles.json'
 
 const mongoUrl = process.env.MONGO_URL || "mongodb://localhost/project-mongo"
 mongoose.connect(mongoUrl, { useNewUrlParser: true, useUnifiedTopology: true })
 mongoose.Promise = Promise
 
+//model for the entries in the database
+const Title = mongoose.model('Title', {
+  show_id: Number,
```

server.js
```diff
@@ -3,35 +3,122 @@ import bodyParser from 'body-parser'
...
+  director: String,
+  cast: String,
+  country: String,
+  date_added: String,
+  release_year: Number,
+  rating: String,
+  duration: String,
+  listed_in: String,
+  description: String,
+  type: String,
```
KarolCislowski, line 25: It will be safer for you if you add `required: true` to some fields. this will prevent from creating empty entries in the database ;)

server.js
```diff
@@ -3,35 +3,122 @@ import bodyParser from 'body-parser'
...
+app.get('/titles', async (req, res) => {
+  const { title, cast, country } = req.query
+  //i means to ignore case
+  const titleRegex = new RegExp(title, 'i')
+  const castRegex = new RegExp(cast, 'i')
+  const countryRegex = new RegExp(country, 'i')
+  //doesn't work when I use this Regex so it only searches for entire words. It works for the queries but it won't find titles anymore.
+  //   const titleRegex = new RegExp('\\b' + title + '\\b', 'i')
+  //   const castRegex = new RegExp('\\b' + cast + '\\b', 'i')
+  //   const countryRegex = new RegExp('\\b' + country + '\\b', 'i')
```
KarolCislowski, line 76: You need to choose if you want to use `//` or `RegExp()` notation.
Here you can see how to use `/b` flag properly:
https://www.w3schools.com/Jsref/jsref_regexp_begin.asp

server.js
```diff
@@ -3,35 +3,122 @@ import bodyParser from 'body-parser'
...
+// app.get('/titles/:show_id', async (req, res) => {
+//   const { show_id } = req.params
+//   const title = await Title.findOne({ show_id: show_id })
+//   if (title) {
+//     res.json(title)
+//   } else {
+//     res.status(404).json({ error: `Could not find title with show id ${show_id}` });
+//   }
+// })
+
```
KarolCislowski, line 100: Please remove unused code from the file :)

server.js
```diff
@@ -3,35 +3,122 @@ import bodyParser from 'body-parser'
...
+//doesn't work, from the codealong
+// app.get('titles/:id', async (req, res) => {
+//   const title = await Title.findById(req.params.id)
+//   if (title) {
+//     res.json(title)
+//   } else {
+//     res.status(404).json({ error: `The title with id: ${id} is not found` })
+//   }
+// })
```
KarolCislowski, line 120: Looks 100% good for me. 
Probably you had `/shows/:show_id` still active while you start to test it, and then express was sending your _id to that route and test it against show_id. Try to comment out `/titles/:id` and uncomment this one while the `titles/:show_id` is not active any more.

PR #54, opened 2020-05-14
Malin's database

malin798:


models.js
```diff
@@ -0,0 +1,12 @@
+import mongoose from 'mongoose'
+
+export const Shows = mongoose.model('Shows', {
+  "show_id": Number,
```
KarolCislowski, line 4: Mongoose adds unique _id field so it is no need to add one by hand. This will make your work harder when you will start adding new records to the database later.

models.js
```diff
@@ -0,0 +1,12 @@
...
+  "title": String,
+  "director": [String],
+  "cast": [String],
+  "country": [String],
+  "release_year": Number,
+  "description": String,
+  "type": String,
```
KarolCislowski, line 11: It will be safer for you if you add `required: true` to some fields. this will prevent from creating empty entries in the database ;)

server.js
```diff
@@ -2,37 +2,162 @@ import express from 'express'
 import bodyParser from 'body-parser'
 import cors from 'cors'
 import mongoose from 'mongoose'
+import showData from './data/netflix-titles.json'
+import { Shows } from './models'
```
KarolCislowski, line 6: Great you extracted models what made your code more organized! 🥇

server.js
```diff
@@ -2,37 +2,162 @@ import express from 'express'
...
 const app = express()
 
-// Add middlewares to enable cors and json body parsing
 app.use(cors())
 app.use(bodyParser.json())
 
-// Start defining your routes here
 app.get('/', (req, res) => {
-  res.send('Hello world')
+  res.sendFile(path.join(__dirname+'/public/index.html'))
```
KarolCislowski, line 23: Serving static file to the user is a great form of saying hello!! Good Job! 👍 
If you want to take another step forward you can try the templating engines like ejs: https://ejs.co

server.js
```diff
@@ -2,37 +2,162 @@ import express from 'express'
...
+  const pageSize = +(req.query.pageSize ?? 20)
+  const startIndex = (page -1) * pageSize
+  const endIndex = startIndex + pageSize 
+  const filteredCategory = categoryFind.slice(startIndex, endIndex)
+  const returnObject = { 
+    pageSize: pageSize,
+    page: page,
+    maxPages: Math.ceil(categoryFind.length/pageSize),
+    totalShows: categoryFind.length,
+    results: filteredCategory
```
KarolCislowski, line 109: You repeated this block of code, so if you still want to deliver pagination this way then the good thing to do would be extracting it to separate function and jus call it when needed :)

server.js
```diff
@@ -2,37 +2,162 @@ import express from 'express'
...
+  const pageSize = +(req.query.pageSize ?? 20)
+  const startIndex = (page -1) * pageSize
+  const endIndex = startIndex + pageSize 
+  const filteredShows = show.slice(startIndex, endIndex)
+  const returnObject = { 
+    pageSize: pageSize,
+    page: page,
+    maxPages: Math.ceil(show.length/pageSize),
+    totalShows: show.length,
+    results: filteredShows
```
KarolCislowski, line 58: You created pagination using the slice method on a very big set of data already fetched from the database. Maybe it would be better to Fetch only part of date we need now from the database. Use mongoose methods like `skip()`, `limit()` and `sort()`.

server.js
```diff
@@ -2,37 +2,162 @@ import express from 'express'
...
+if (process.env.RESET_DB) {
+	const seedDatabase = async () => {
+    await Shows.deleteMany({})
+
+		showData.forEach((show) => {
+      const splitCast = show.cast.split(", ")
+      show.cast = splitCast
+
+      if (show.director === "") {
+        show.director = "Unknown"
```
KarolCislowski, line 141: Nice, you spotted and handled the empty strings. 👍

PR #55, opened 2020-05-14
Project API Mongoose - Mathias T - Review wanted

Nesiht:
Fun project, had to redo all when I realized all my filter methods was using the .json dataset and not the data stored in mongoose 😄 
Mongoose documentation was really good, read a lot and it helped much. 

[Heroku](https://michelin-restaurants.herokuapp.com/)


server.js
```diff
@@ -2,37 +2,134 @@ import express from 'express'
...
+    res.status(404).json({ message: `Error, no restaurants found in ${city}` })
+  }
+  res.json({ totalResults: allUniqueCitys.length, allUniqueCitys })
+})
+
+// Find restaurant by name
+// Use parameter with /michelin/pierre
+app.get('/michelin/:name', async ( req, res ) => {
+  const { name } = req.params
+  const restaurantByName = await Michelin.findOne({ name: name })
```
HIPPIEKICK, line 111: Good that you found the `findOne` method! Downside is that it's just returning the first match, so if two restaurants have the same name, maybe this doesn't return what I want. I think you should include the name search as a query param instead, and use `findOne` for something you know is unique - like the id.

server.js
```diff
@@ -2,37 +2,134 @@ import express from 'express'
...
+app.get('/michelin/year/:year', async ( req, res ) => {
+  const { year } = req.params
+  const restaurantByYear = await Michelin.find({ year: year })
+
+  if ( restaurantByYear.length > 0 ) {
+    res.json({ totalResults: restaurantByYear.length, restaurantByYear})
+  } else {
+    res.status(404).json({ message: `Error, no restaurants by that year` })
+  }
+})
```
HIPPIEKICK, line 130: One advice to make it more RESTful is that you name your main endpoint descriptive in plural, for example `/restaurants`, and then you add queries for the parameters. The endpoints should be descriptive of what it returns and not how they are filtered (meaning, this endpoint doesn't return years, it filters on years). So you could have queries like this:
```
/restaurant?region=YOUR_COUNTRY
/restaurant?cuisine=YOUR_CUISINE
/restaurant?city=YOUR_CITY
/restaurant?name=SEARCH_NAME
/restaurant?year=SEARCH_YEAR

```

PR #58, opened 2020-05-15
Project Mongo API - RAGNA

Ragtulf:
I feel like I totally misunderstood this week's objective and, by chance, I  asked Van about something unrelated when he commented that this week we were supposed to move away from using javascript to filter the data we had fetched from the API. I don't have any more time to fix this project or learn what I need because I need to move on to next week's project. I'm frustrated that I didn't get the object of the week.

Regular Expressions. And how to use the filters available in mongoose? I don't even know what I'm asking.


server.js
```diff
@@ -23,16 +39,112 @@ mongoose.Promise = Promise
...
+    yearNominations = yearNominations.filter((item) => item.win === true)
+  }
+  res.json(yearNominations)
+
+  if (yearNominations) {
+    res.json(yearNominations)
+  } else {
+    res.status(404).json({ error: 'Not found' })
+  }
+})
```
KarolCislowski, line 102: Here you are fetching a big set of data no matter if you want only the winners or not.
I think a better option would be to check if win=true and then use `Nomination.find()` with or without win parameter to not load your servers with extra computing and transfers?

server.js
```diff
@@ -23,16 +39,112 @@ mongoose.Promise = Promise
...
+    let wonCategories = categories.filter((item) => item.win === true)
+    wonCategories = wonCategories.filter((item) => item.year_award === +year)
+    res.json(wonCategories)
+  } else if (categories && year) {
+    const yearCategories = categories.filter((item) => item.year_award === +year)
+    res.json(yearCategories)
+  } else {
+    res.status(404).json({ error: 'Not found' })
+  }
 })
```
KarolCislowski, line 145: Here again. Filtering by using mongoose is much easier. You just put all criteria to the find() method.
You can read more here: https://mongoosejs.com/docs/api.html#model_Model.find
And if you still will have a lot of questions just ping me on Slack and we will try to make it clear for you.

PR #59, opened 2020-05-15
Jihyun An project-mongo-api

jihyeona:
I used golden globes nomination data and I had problem using RegExp for the query to filter the nominations that won. I think this was because the value of this query(win) is boolean, and I couldn't figure out how to pass in boolean value in ReqExp. So I skipped using RegExp and instead used .where() to filter the results based on the key 'win'. Maybe there is a better solution to this? 
I also skipped using RegExp for pagination because I couldn't see a point to use it. It is bit confusing for me still to understand where to use RegExp and why it makes sense to use them.


jihyeona:
Hi! Thank you so much for a detailed review! It really helps. Now I've made some changes based on the review including a route to return a single result. 

I have a question about using multiple queries though. With my code it only works to use one single query at a time since it is checking if a query is there, and then going to the next else if statement. I can't think of a smart way to write a code so that it would be possible to use multiple queries at the same time.. do you have any suggestion or tips on this?

HIPPIEKICK:
> Hi! Thank you so much for a detailed review! It really helps. Now I've made some changes based on the review including a route to return a single result.
> 
> I have a question about using multiple queries though. With my code it only works to use one single query at a time since it is checking if a query is there, and then going to the next else if statement. I can't think of a smart way to write a code so that it would be possible to use multiple queries at the same time.. do you have any suggestion or tips on this?

Good job! You could do like this:
```
const { nominee, category, win, year } = req.query
const nominations = await Nomination.find({ 
  nominee: new RegExp(nominee, 'i'),
  category: new RegExp(category, 'i'),
  //etc...
})
```

PR #60, opened 2020-05-15
Project Mongo API Lisa Sturegård

LisaStur:
Very interesting week! Had some issues in getting the deployed site to connect with the database, but got it working thanks to help from Van. So good to learn how databases and endpoints work and how to create them! Super useful regardless if you work with front end or some other part of the chain!
https://db-learning-week18.herokuapp.com/


models/show.js
```diff
@@ -0,0 +1,18 @@
...
+  director: { type: String },
+  cast: { type: String },
+  country: { type: String },
+  date_added: { type: String },
+  release_year: { type: Number },
+  rating: { type: String },
+  duration: { type: String },
+  listed_in: { type: String },
+  description: { type: String },
+  type: { type: String },
```
HIPPIEKICK, line 15: Nice that you extracted this to its own file! If you just want to specify the type, you could write it in one line:
```
show_id: Number,
title: String,
```
etcetera, and if you have more attributes, you write like this:
```
show_id: {
  type: Number,
  unique: true,
},
title: {
  type: String
  required: true,
},

```

server.js
```diff
@@ -2,37 +2,82 @@ import express from 'express'
...
+if (process.env.RESET_DB) {
+  console.log('Resetting Database!')
+  const seedDatabase = async () => {
+    await NetflixShow.deleteMany({})
+    netflixData.forEach((show) => {
+      new NetflixShow(show).save()
+    })
+  }
+  seedDatabase()
+}
```
HIPPIEKICK, line 21: Nice!

server.js
```diff
@@ -2,37 +2,82 @@ import express from 'express'
...
-// Add middlewares to enable cors and json body parsing
 app.use(cors())
 app.use(bodyParser.json())
+app.use((req, res, next) => {
+  if (mongoose.connection.readyState === 1) {
+    next()
+  } else {
+    res.status(503).json({ error: 'Service unavailable'})
+  }
+})
```
HIPPIEKICK, line 34: Good check and clear error messsage!

server.js
```diff
@@ -2,37 +2,82 @@ import express from 'express'
...
 
-// Start defining your routes here
 app.get('/', (req, res) => {
-  res.send('Hello world')
+  res.send(`<p>Netflix show data with endpoints: <br> /netflixshows<br> /netflixshows?query="Your Query"<br> /netflixshows/type?type="Movie/TV Show"<br> /netflixshows/rating?rating="TV-Y/TV-Y7/TV-Y7-FV/TV-G/G/TV-PG/PG/PG-13/TV-14/TV-MA/R"<br> /netflixshows/title/"Your Title"<p>`)
```
HIPPIEKICK, line 37: For nice formatting - take a look at this: [https://www.npmjs.com/package/express-list-endpoints](https://www.npmjs.com/package/express-list-endpoints)

server.js
```diff
@@ -2,37 +2,82 @@ import express from 'express'
...
+})
+
+app.get('/netflixshows', async (req, res) => {
```
HIPPIEKICK, line 40: Descriptive and good naming!

server.js
```diff
@@ -2,37 +2,82 @@ import express from 'express'
...
+  const { query } = req.query
+  const queryRegEx = new RegExp(query, 'i') 
+  const netflixShows = await NetflixShow.find({ title: queryRegEx }).sort({title: 1})
+  res.json(netflixShows)
+})
```
HIPPIEKICK, line 45: Since you're only searching the title, maybe a more descriptive name for the query would be title? `const { title } = req.query`

server.js
```diff
@@ -2,37 +2,82 @@ import express from 'express'
...
+  const { type } = req.query
+  let shows = await NetflixShow.find()
+  if (type) { 
+    shows = shows.filter((show) => show.type.toString() === type)
+  }
+  if (shows.length === 0) {
+    res.status(404).send('Please select Movie or TV Show to filter on "type"')
+  }
+  res.json(shows)
+})
```
HIPPIEKICK, line 57: It would be more RESTful if this query was in the `/netflixshows` endpoint. And when working with mongo, try to stick to mongo methods like `find`, instead of regular javascript methods like `filter()`. (And maybe add regex for case insensitive here to make it possible to write in lower case)

server.js
```diff
@@ -2,37 +2,82 @@ import express from 'express'
...
+  const { rating } = req.query
+  let shows = await NetflixShow.find()
+  if (rating) {
+    shows = shows.filter((show) => show.rating.toString() === rating)
+  }
+  if (shows.length === 0) {
+    res.status(404).send('Available ratings: TV-Y, TV-Y7, TV-Y7-FV, TV-G, G, TV-PG, PG,  PG-13, TV-14, TV-MA and R')
+  }
+  res.json(shows) 
+})
```
HIPPIEKICK, line 69: Same with this one, it would be more RESTful if this query was in the `/netflixshows` endpoint. And use mongo methods

server.js
```diff
@@ -2,37 +2,82 @@ import express from 'express'
...
+
+app.get('/netflixshows/title/:title', async (req, res) => {
+  const { title } = req.params
+  const show = await NetflixShow.findOne({ title })
```
HIPPIEKICK, line 73: Super nice that you're using the `findOne` method! Downside is that it's just returning the first match, so if two shows have the same name, maybe this doesn't return what I want. I think you should include the title search as a query param for the `/netflixshows` endpoint instead, and use `findOne` for something you know is unique - like the `show_id`.

server.js
```diff
@@ -2,37 +2,82 @@ import express from 'express'
...
+  if (show) {
+    res.json(show)
+  } else {
+    res.status(404).json({ error: `Could not find ${title}` })
```
HIPPIEKICK, line 77: ⭐

server.js
LisaStur: Super, will give it a try, big thanks! :)

PR #61, opened 2020-05-15
Week 18 - ebbabw_project-mongo-api

ebbabw:
Link to our deployed project:
https://ebbbw-project-mongo-api.herokuapp.com/

Our reflections on how this week's project turned out:
It didn't go so well this time but I'm happy that I have something 

Things we'd like to have clarified or explained in more detail:

Tried to return a singel netflixtitle id but was overrinding my other code 
I managed to return (A list of directors) but was override by the ID function 

It's now commented out (returning one id, it worked without list of directors)
Couldn't find the right way to ask the question on stack.

### DIFF
--- a/server.js
+++ b/server.js
@@ -64,7 +64,7 @@ app.use((req, res, next) => {
 // Start defining your routes here
 
 app.get('/', (req, res) => {
-  res.send('Hello! use these routes /netflixtitles (list all netflixtitles) /netflixtitles/directors (list all directors)')
+  res.send('Hello! use these routes /netflixtitles (list all netflixtitles)  /netflixtitles/:id (try 5ec44053d9b97d0023f16da1") /directors (list all directors)')
 })
 
 app.get("/netflixtitles", async (req, res) => {
--- a/server.js
+++ b/server.js
@@ -16,6 +16,7 @@ mongoose.Promise = Promise
 const Netflixdata = mongoose.model('Netflixdata', {
 
   title: { type: String }, 
+  director: { type: String },
   cast:  { type: String },
   country: { type: String },
   date_added: { type: String },
@@ -29,21 +30,10 @@ const Netflixdata = mongoose.model('Netflixdata', {
 
 const Director = mongoose.model('Director', {
 
-  director: { type: String },
+  director: { type: String},
+
 
 })
-
-
-// const Netflixtitle = mongoose.model('Netflixtitle', {
-
-//   director: String,
-
-//   title: {
-//     type: mongoose.Schema.Types.ObjectId,
-//     ref: 'Title'
-//   }
-
-// });
 
 if (process.env.RESET_DATABASE) {
   console.log("Resetting database...");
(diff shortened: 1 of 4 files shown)


server.js
```diff
@@ -3,36 +3,125 @@ import bodyParser from 'body-parser'
...
 const mongoUrl = process.env.MONGO_URL || "mongodb://localhost/project-mongo"
-mongoose.connect(mongoUrl, { useNewUrlParser: true, useUnifiedTopology: true })
+mongoose.connect(Key, { useNewUrlParser: true, useUnifiedTopology: true })
 mongoose.Promise = Promise
 
-// Defines the port the app will run on. Defaults to 8080, but can be 
-// overridden when starting the server. For example:
-//
-//   PORT=9000 npm start
+const Netflixdata = mongoose.model('Netflixdata', {
```
HIPPIEKICK, line 16: Your model should have the singular name of the collection your model is for.
`const NetflixTitle = mongoose.model('NetflixTitle', {`

server.js
```diff
@@ -3,36 +3,125 @@ import bodyParser from 'body-parser'
...
+  director: { type: String },
+  cast:  { type: String },
+  country: { type: String },
+  date_added: { type: String },
+  release_year: { type: Number },
+  rating: { type: String },
+  duration: { type: String },
+  listed_in: { type: String },
+  description: { type: String },
+  type: { type: String }
```
HIPPIEKICK, line 28: When you only specify type you could write it on one line:
```
title: String, 
director: String,
```
etcetera, but when you start adding other properties you need to do like you've done:
```
title: {
  type: String,
  required: true
},
director: {
  type: String,
  required: true
},
```

server.js
```diff
@@ -3,36 +3,125 @@ import bodyParser from 'body-parser'
...
 app.use(cors())
 app.use(bodyParser.json())
 
+app.use((req, res, next) => {
+  if (mongoose.connection.readyState === 1) {
+    next()
+  } else {
+    res.status(503).json({ error: 'Service unavailable' })
+  }
+})
```
HIPPIEKICK, line 63: Nice check and clear error message!

server.js
```diff
@@ -3,36 +3,125 @@ import bodyParser from 'body-parser'
...
 })
 
+app.get("/netflixtitles", async (req, res) => {
+  const { query } = req.query
+  const queryRegex = new RegExp(query, 'i');
+  const titles = await Netflixdata.find({title: queryRegex}).sort({
+    release_year: ""
+  })
+  res.json(titles.reverse());
+});
```
HIPPIEKICK, line 78: Since you're only searching for title, maybe your query param could be called title instead of query?

server.js
```diff
@@ -3,36 +3,125 @@ import bodyParser from 'body-parser'
...
 // Start defining your routes here
+
 app.get('/', (req, res) => {
-  res.send('Hello world')
+  res.send('Hello! use these routes /netflixtitles (list all netflixtitles) /netflixtitles/directors (list all directors)')
 })
 
+app.get("/netflixtitles", async (req, res) => {
+  const { query } = req.query
+  const queryRegex = new RegExp(query, 'i');
```
HIPPIEKICK, line 73: ⭐

server.js
```diff
@@ -3,36 +3,125 @@ import bodyParser from 'body-parser'
...
+  } else {
+    res.status(404).json({ error: 'error could not found' });
+  }
+
+} catch(err) {
+ res.status(400).json({error: 'Something is Invalid'})
+
+}
+
+});
```
HIPPIEKICK, line 115: Yeah, as you said, one endpoint gets overwritten when you have the two path params at the same endpoint. Why not change the name of this endpoint, since endpoints should be named after what they return - which in this case is directors `/directors`

server.js
```diff
@@ -3,36 +3,125 @@ import bodyParser from 'body-parser'
...
-// import topMusicData from './data/top-music.json'
+//API input
+import netflixTitles from './data/netflix-titles.json'
+
+
+const Key = 'mongodb+srv://ebbabw:gs4m4g4U6pC6ejLK@cluster0-yp7wy.mongodb.net/project-mongo?retryWrites=true&w=majority'
 
 const mongoUrl = process.env.MONGO_URL || "mongodb://localhost/project-mongo"
-mongoose.connect(mongoUrl, { useNewUrlParser: true, useUnifiedTopology: true })
+mongoose.connect(Key, { useNewUrlParser: true, useUnifiedTopology: true })
```
HIPPIEKICK, line 13: You should keep your mongoUrl as an environmental variable and not write it out like this.

ebbabw:
Thank you for all the inputs and corrections! I have now update the app so I think everything is working as it should 👍

HIPPIEKICK:
> Thank you for all the inputs and corrections! I have now update the app so I think everything is working as it should 👍

I took a look at your heroku app now, and it says "error": "Service unavailable". I saw that you removed the connection string to mongo from the code - just want to make sure that you added the `MONGO_URL` as a config var in heroku? Can you double check that the connection string is correct?

ebbabw:
Now it works again, I had changed the password on the new project and forgot to update to the new in heroku, Now it should work again :)

HIPPIEKICK:
Super!

PR #66, opened 2020-05-15
Olle Alm - mongoDB Books

ollealm:
A bit tricky to know how to think about DB data at first, but I think I got it after a while.

I spent a lot of time trying to solve how to seed every unique author only once and then populate the books with the correct author. I did get it to work last minute, and would love to get a comment if it's done in an ok way.

https://olle-mongodb-books.herokuapp.com/

### DIFF
--- a/server.js
+++ b/server.js
@@ -12,7 +12,7 @@ import mongoose from 'mongoose'
 // import netflixData from './data/netflix-titles.json'
 // import topMusicData from './data/top-music.json'
 
-const mongoUrl = process.env.MONGO_URL || "mongodb://localhost/project-mongo"
+const mongoUrl = process.env.MONGO_URL || "mongodb://localhost/books"
 mongoose.connect(mongoUrl, { useNewUrlParser: true, useUnifiedTopology: true })
 mongoose.Promise = Promise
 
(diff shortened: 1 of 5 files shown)


server.js
```diff
@@ -27,11 +22,105 @@ const app = express()
...
-// Start defining your routes here
+
+if (process.env.RESET_DB) {
+  console.log('Database reset')
+  const seedDatabase = async () => {
+    await Author.deleteMany({})
+    await Book.deleteMany({})
+
+    // creating array with unique Authors from data
+    const uniqueAuthors = [... new Set(booksData.map(book => book.authors))]
```
puj, line 33: This is a really good find to prevent duplicate authors from being present in your seed data. Nice

server.js
```diff
@@ -27,11 +22,105 @@ const app = express()
...
+          title: book.title,
+          author: author, // unique author
+          average_rating: book.average_rating,
+          isbn: book.isbn,
+          isbn13: book.isbn13,
+          num_pages: book.num_pages,
+        }).save()
+      })
+    })
+  }
```
puj, line 52: Nice, find the books by this author and save them a this point.  A nice thought exercise would be to think "What if authors was an array of different authors"  How would I prevent duplicate books?

server.js
```diff
@@ -27,11 +22,105 @@ const app = express()
...
+      .populate('author')
+    res.json(books)
+  } else {
+    const books = await Book.find().sort(sortQuery(sort)).populate('author')
+    res.json(books)
+  }
 })
 
+app.get('/books/:id', async (req, res) => {
+  const book = await Book.findById(req.params.id).populate('author')
```
puj, line 88: Good

server.js
```diff
@@ -27,11 +22,105 @@ const app = express()
...
+app.get('/books', async (req, res) => {
+  const { title, sort } = req.query
+  console.log(req.query)
+  const titleRegex = new RegExp(`\\b${title}\\b`, 'i')
+
+  const sortQuery = (sort) => {
+    if (sort === 'rating') return { average_rating: -1 }
+    if (sort === 'pages') return { num_pages: -1 }
+    if (sort === 'title') return { title: 1 }
+  }
```
puj, line 72: Cool approach.  Good understanding of what Mongoose is doing.

PR #68, opened 2020-05-16
Done for review, plz

Tuliany:


server.js
```diff
@@ -1,35 +1,128 @@
...
 
 // Add middlewares to enable cors and json body parsing
 app.use(cors())
 app.use(bodyParser.json())
 
 // Start defining your routes here
+
 app.get('/', (req, res) => {
-  res.send('Hello world')
+  res.send('My endpoints= /shows & /shows?query=movie or /shows?query=Tv Show & /shows/:title & /shows/year/:release_year & /shows/country/:country' )
```
HIPPIEKICK, line 84: Nice with documentation!

server.js
```diff
@@ -1,35 +1,128 @@
...
+})
+
+app.get('/shows', async (req, res) => {
+  const { query } = req.query
```
HIPPIEKICK, line 88: I thought this was a general query so I tried to search for chocolate 😎 Can you name it something more descriptive so that the user knows this query is searching for the type of show?

server.js
```diff
@@ -1,35 +1,128 @@
...
+   if (query) {
+    //localhost:5000/shows?query=Showtype 
+    const queryRegex = new RegExp(query, 'i')
```
HIPPIEKICK, line 91: Nice with regex!

server.js
```diff
@@ -1,35 +1,128 @@
...
+    const shows = await Show.find({ type: queryRegex })
+    res.json(shows)
+   } else {
+    //localhost:5000/shows
+    const shows = await Show.find();
+    res.json(shows)
```
HIPPIEKICK, line 97: Looks a bit weird to me to define the same `const` twice (even though it works), and a little repetitive with two `res.json`. If you define it as a `let` before the if-statement you could reassign it in the if-statement. Example:
```
let shows = await Show.find()
if (query) {
  const queryRegex = new RegExp(query, 'i')
  shows = await Show.find({ type: queryRegex })
}
res.json(shows)
```

server.js
```diff
@@ -1,35 +1,128 @@
...
+
+app.get('/shows/:title', async (req, res) => {
+  const { title } = req.params
+  const show = await Show.findOne({ title: title })
+  if (show) {
+    res.json(show)
+  } else {
+    res.status(404).json({ error: `Could not find show with title=${title}` })
+  }
+})
```
HIPPIEKICK, line 109: `findOne` will only give you the first match, so if two shows has the same name - maybe the wrong one will be returned. Mongoose gives you a unique `_id` which you can use for the endpoint that only should return a single result.

server.js
```diff
@@ -1,35 +1,128 @@
...
+    res.json(year)
+  } else {
+    res.status(404).json({ error: 'Not found' })
+  }
+})
+
+app.get('/shows/country/:country', async (req, res) => {
+  const { country } = req.params
+  const whatCountry = await Show.find({ country: country })
+    res.json(whatCountry)
```
HIPPIEKICK, line 124: Endpoints should be named after the data they are returning, and not the way they are filtering the data. To make it more RESTful you could include these two as queries in you `/shows` endpoint:
`/shows?year=2010&country=sweden`

server.js
```diff
@@ -1,35 +1,128 @@
 import express from 'express'
 import bodyParser from 'body-parser'
 import cors from 'cors'
-import mongoose from 'mongoose'
+import mongoose, { MongooseDocument } from 'mongoose'
+import { stringify } from 'querystring'
```
HIPPIEKICK, line 5: Are you using this?

server.js
```diff
@@ -1,35 +1,128 @@
...
+   }
+ })
+
+app.get('/shows/:title', async (req, res) => {
+  const { title } = req.params
+  const show = await Show.findOne({ title: title })
+  if (show) {
+    res.json(show)
+  } else {
+    res.status(404).json({ error: `Could not find show with title=${title}` })
```
HIPPIEKICK, line 107: Descriptive ⭐

PR #69, opened 2020-05-16
mongo-api-project

louiseeriksson:


HIPPIEKICK:
Hi Louise! On your deployed site I get `"error": "Service unavailable"`. Can you take a look at Heroku (make sure the MONGO_URL has the right connection link). If it looks fine, you can also check heroku's logs by going in to your application in heroku and then clicking "More" in the top right corner, and then "View logs". Hopefully this will give you a clue what's causing this.

louiseeriksson:
I've checked the MONGO_URL and Heroku logs.. don't understand whats wrong. I'll schedule a 1:1 talk with a teacher or see if someone in my team can help me

server.js
```diff
@@ -2,36 +2,84 @@ import express from 'express'
...
+//Find by title: http://localhost:8080/title/Titanic
+app.get('/title/:title', (req, res) => {
+  Movie.findOne({ title: req.params.title }).then(movie => {
+    if (movie) {
+      res.json(movie)
+    } else {
+      res.status(404).json({ error: 'Movie not found!' })
+    }
+  })
+})
```
HIPPIEKICK, line 61: Endpoints' names should be in plural (and then the path param could be in singular): `/titles/:title`.

What happens if two movies has the same name? `findOne` just returns the first match. Consider searching for something unique, like `id` instead.

server.js
```diff
@@ -2,36 +2,84 @@ import express from 'express'
...
+
+
+// Find by type: http://localhost:8080/multiple/?type=Movie
+// type & release_year: http://localhost:8080/multiple?type=Movie&release_year=2019
+// type & release_year & rating: http://localhost:8080/multiple?type=Movie&release_year=2019&rating=TV-G
```
HIPPIEKICK, line 66: These example endpoints could be placed in the readme instead. There's also a nice npm package that you could use for documenting your endpoints in a nice way: [https://www.npmjs.com/package/express-list-endpoints](https://www.npmjs.com/package/express-list-endpoints)

server.js
```diff
@@ -2,36 +2,84 @@ import express from 'express'
...
+    type: new RegExp(req.query.type, 'i'),
+    rating: new RegExp(req.query.rating, 'i'),
+  }).then(movies => {
+    if (movies.length > 0) {
+      res.json(movies)
+    } else {
+      res.status(404).json({ error: "Couldn't find a match for your filter. Try again!" })
+    }
+  })
+})
```
HIPPIEKICK, line 80: These queries could be placed under the `/titles` endpoint (the one you named `'/'`), to make it more restful. Endpoints should be named after what they are returning (`/titles`) and not how they are filtered (`/multiple`)

server.js
```diff
@@ -2,36 +2,84 @@ import express from 'express'
...
+  rating: String
+})
+
+const seedDatabase = async () => {
+  await Movie.deleteMany()
+  netflixData.forEach((movie) => new Movie(movie).save())
+
+  new Movie({ title: 'Titanic', release_year: 1997, type: 'Movie' }).save()
+  new Movie({ title: 'Sons of Anarchy', release_year: 2008, type: 'TV Show' }).save()
+  new Movie({ title: 'Forrest Gump', release_year: 1994, type: 'Movie' }).save()
```
HIPPIEKICK, line 24: I agree, these should definitely be on Netflix 😎

PR #70, opened 2020-05-17
Week 18 Project Mongo API

VickBe:
It went ok. 
Could have done more endpoints and stretch goals but ran out of time as I had to work a lot this week :(
I made queries.

### DIFF
--- a/server.js
+++ b/server.js
@@ -4,14 +4,6 @@ import cors from 'cors'
 import mongoose from 'mongoose'
 import goldenGlobesData from './data/golden-globes.json'
 
-// If you're using one of our datasets, uncomment the appropriate import below
-// to get started!
-// 
-// import goldenGlobesData from './data/golden-globes.json'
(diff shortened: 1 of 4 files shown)


HIPPIEKICK:
Looks like your deployed version has no data 😕Did you add `RESET_DATABASE` as a config var in heroku to seed your database?

server.js
```diff
@@ -2,37 +2,121 @@ import express from 'express'
...
+  // })
+  
+
+
+
+//GOLDEN GLOBES FILTERED ON ID - RETURNS A SINGLE RESULT
+app.get('/goldenGlobes/id/:id', async (req, res) => {
+  const goldenGlobes = await goldenGlobes.findById(req.params.id)
+  res.json(nominations)
+})
```
HIPPIEKICK, line 117: This endpoint gives application error. The `res.json` doesn't match the variable name, so if you change that it should be fine!

PR #75, opened 2020-06-09
Forgotten PR. Emelies Project-mongo-api

emlo2018:


server.js
```diff
@@ -2,36 +2,103 @@ import express from 'express'
...
+      if (nominee) {
+        res.json(nominee)
+      } else {
+        res.status(404).json(`error: Nominee ${nominee} not found`)
+      }
+    } catch (err) {
+    res.status(400).json(`Bad Request: this error is an HTTP status code that means that the request you sent to the website server, often something simple like a request to load a web page, was somehow incorrect or corrupted and the server couldn't understand it.`)
+    }
+  })
+})
```
HIPPIEKICK, line 77: To make it more RESTFUL, you could divide it like this:
`GET /nominees/:nomineeID` - search for ID for one specific nominee (because, what happens if two nominees have the same name?
`GET /nominees` - and add a query for name searches: `/nominees?name=avatar`

server.js
```diff
@@ -2,36 +2,103 @@ import express from 'express'
...
+        res.status(404).json(`error: Nominee ${nominee} not found`)
+      }
+    } catch (err) {
+    res.status(400).json(`Bad Request: this error is an HTTP status code that means that the request you sent to the website server, often something simple like a request to load a web page, was somehow incorrect or corrupted and the server couldn't understand it.`)
+    }
+  })
+}) 
+
+/* All nominations for specific year it was nominated and if winner*/
+app.get('/year/:year', (req, res) => {
```
HIPPIEKICK, line 80: Endpoints should be named after what they return and not how they are filtered, so to make it more RESTful, this could be a query under your `/nominee`s endpoint instead:
`/nominees?year=2010`

server.js
```diff
@@ -2,36 +2,103 @@ import express from 'express'
...
+
+app.get('/winners', (_, res) => {
+
+  let winners = globeData.filter(item => {
+    return item.win === true
+  })
+  
+  res.json(winners)
+
+})
```
HIPPIEKICK, line 99: This endpoint doesn't work, since you already have `/:nominee`. So if I write `/winners`, it will think it's a name search for a nominee called "winners"

server.js
```diff
@@ -2,36 +2,103 @@ import express from 'express'
...
+
+app.get('/:nominee', (req, res) => {
+
+  Nominee.findOne({ nominee: req.params.nominee }).then(nominee => {
+    try {
+      if (nominee) {
+        res.json(nominee)
+      } else {
+        res.status(404).json(`error: Nominee ${nominee} not found`)
```
HIPPIEKICK, line 71: `else` will kick in `if !nominee`, therefore `${nominee}` = null here

PR #78, opened 2020-12-15
Olofs Mongo API

luuf81:
Since I'm leaving for a ski trip on Wednesday I'm making this pull request quite early and not with all the functionality I would probably have implemented if I had more time :)

### DIFF
--- a/server.js
+++ b/server.js
@@ -73,9 +73,25 @@ app.get("/shows", async (req, res) => {
   res.json(shows);
 });
 
+//Get one show by showid
+
+app.get('/shows/:id', async (req, res) => {
+  try {
+  console.log(req.params)
+  const showId = await Show.findOne({ show_id: req.params.id })
+  if(showId) {
+    res.json(showId)
+  } else {
+    res.status(404).json({ error: 'Show not found' })
+  }
+} catch (err) {
+  res.status(400).json({ error: 'Invalid show id' })
+}
+})
+
 //shows limited to 10 documents, paginated with query "page"
 
-app.get("/shows/limited", async (req, res) => {
+app.get("/shows", async (req, res) => {
   const page = req.query.page;
   const shows = await Show.find()
     .skip((page === 0 ? page : 1) * 10 - 10) //page value 0 will resort to 1 since monbgoDB doesn't accept otherwise. Strangely this works locally but not deployed to heroku...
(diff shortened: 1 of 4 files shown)


server.js
```diff
@@ -1,38 +1,97 @@
...
+//shows limited to 10 documents, paginated with query "page"
+
+app.get("/shows/limited", async (req, res) => {
+  const page = req.query.page;
+  const shows = await Show.find()
+    .skip((page === 0 ? page : 1) * 10 - 10) //page value 0 will resort to 1 since monbgoDB doesn't accept otherwise. Strangely this works locally but not deployed to heroku...
+    .limit(10);
+  res.json(shows);
+});
+
```
LillanEriksson, line 85: Nice with the pagination! Hope you'll get it to work in the deployed version also

LillanEriksson (approved):
Nice work! Nothing to comment on really.

server.js
```diff
@@ -1,38 +1,97 @@
...
+//shows limited to 10 documents, paginated with query "page"
+
+app.get("/shows/limited", async (req, res) => {
+  const page = req.query.page;
+  const shows = await Show.find()
+    .skip((page === 0 ? page : 1) * 10 - 10) //page value 0 will resort to 1 since monbgoDB doesn't accept otherwise. Strangely this works locally but not deployed to heroku...
+    .limit(10);
+  res.json(shows);
+});
```
KarolCislowski, line 84: Great that you made this effort to create this functionality, but when you are using queries to sort/filter the data set, do not create a separate endpoint. You should do this in `/shows` endpoint. 

A separate endpoint is good if you want to use props like this `/shows/limited/:page`

server.js
```diff
@@ -1,38 +1,97 @@
...
+
+//parameter country
+
+app.get("/shows/:country", async (req, res) => {
+  const { country } = req.params;
+  const shows = await Show.find({ country: country });
+  res.json(shows);
+});
```
KarolCislowski, line 92: When you're using params to filter the data you need to inform the user what field is the key to this filter.

`/shows/:id` - because it is the default key.
`/shows/countries/:country` - now user knows that he/she is looking for country.

KarolCislowski (changes requested):
Hey very nice work, but to meet the blue level you need to create an endpoint which will return only 1 document from your database.

server.js
luuf81: Thanks. changed this to your suggestion.

luuf81:
Strangely I missed the endpoint with a single document from the database. That is now fixed!

KarolCislowski (approved):
Great Work!! 👍🏻

PR #81, opened 2020-12-16
moa hammarbäck - mongodb api

mhammarback:
This week i aimed for the blue level requirements.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,7 @@
 # Mongo API Project
-This project was made during the Technigo Bootcamp. The goal was to practice how to use Mongodb to store data and how to query that data from an API.  
+This project was made during the Technigo Bootcamp. The goal was to practice how to use Mongodb to store data and how to query that data from an API using mongoose operations.  
 
-## The problem
+## What i've learned
 
 - I've created endpoints and i've used mongoose operations instead of vanilla JS since the data needed to be manipulated.
 - I've learned how to use the mongoose methods find() and findOne().
--- a/README.md
+++ b/README.md
@@ -1,13 +1,21 @@
 # Mongo API Project
-
-Replace this readme with your own information about your project. 
-
-Start by briefly describing the assignment in a sentence or two. Keep it short and to the point.
+This project was made during the Technigo Bootcamp.  The goal was to practice how to use Mongodb to store data and how to query that data from an API.  
 
 ## The problem
 
-Describe how you approached to problem, and what tools and techniques you used to solve it. How did you plan? What technologies did you use? If you had more time, what would be next?
+- I've created endpoints and i've used mongoose operations instead of vanilla JS since the data needed to be manipulated.
+- I've learned how to use the mongoose methods find() and findOne().
+
+## Routes 
+
+GET /welcome 
+
+GET /books - displays 500 books
+
+GET /books/book/:bookID - displays a single book based on the bookID parameter. Error message will show if the bookID is invalid.
+
+GET /books/authors/:authorName  - displays books by a specific author. I've added Regex so that the search will be non-case-sensitive and and Error will show if no author is found.  
+
 
 ## View it live
 
-Every project should be deployed somewhere. Be sure to include the link to the deployed project so that the viewer can click around and see what it's all about.
--- a/README.md
+++ b/README.md
@@ -1,10 +1,11 @@
 # Mongo API Project
-This project was made during the Technigo Bootcamp.  The goal was to practice how to use Mongodb to store data and how to query that data from an API.  
+This project was made during the Technigo Bootcamp. The goal was to practice how to use Mongodb to store data and how to query that data from an API.  
 
 ## The problem
 
 - I've created endpoints and i've used mongoose operations instead of vanilla JS since the data needed to be manipulated.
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -1,13 +1,22 @@
...
 
-Describe how you approached to problem, and what tools and techniques you used to solve it. How did you plan? What technologies did you use? If you had more time, what would be next?
+GET /welcome 
 
-## View it live
+GET /books - displays 500 books
+
+GET /books/book/:bookID - displays a single book based on the bookID parameter. Error message will show if the bookID is invalid.
+
+GET /books/authors/:authorName  - displays books by a specific author. I've added Regex so that the search will be non-case-sensitive and and Error will show if no author is found.
```
HenrikeW, line 19: nice that you describe your paths here!

README.md
```diff
@@ -1,13 +1,22 @@
...
-Start by briefly describing the assignment in a sentence or two. Keep it short and to the point.
+- I've created endpoints and i've used mongoose operations instead of vanilla JS since the data needed to be manipulated.
+- I've learned how to use the mongoose methods find() and findOne().
+- I've created an account on Mongodb Atlas and succesfully manage to store my data. 
 
-## The problem
+## Routes 
 
-Describe how you approached to problem, and what tools and techniques you used to solve it. How did you plan? What technologies did you use? If you had more time, what would be next?
+GET /welcome
```
HenrikeW, line 12: this path doesn't work unfortunately. Maybe you just meant `GET /` ?

server.js
```diff
@@ -3,36 +3,71 @@ import bodyParser from 'body-parser'
...
+if (process.env.RESET_DB) {
+  const seedDatabase = async () => {
+    await Book.deleteMany()
+
+    booksData.forEach((item) => {
+      new Book(item).save()
+    }) 
+  }
+  seedDatabase()
+}
```
HenrikeW, line 34: clean & easy 👍

server.js
```diff
@@ -3,36 +3,71 @@ import bodyParser from 'body-parser'
...
+  res.send('Welcome!')
+})
+
+app.get('/books', async ( req, res) => {
+  const allBooks = await Book.find()
+  res.json(allBooks)
+})
+
+app.get('/books/book/:bookID', async (req, res) => {
+  const oneBook = await Book.findOne( { bookID: req.params.bookID })
```
HenrikeW, line 52: nice to solve this in just one line!

server.js
```diff
@@ -3,36 +3,71 @@ import bodyParser from 'body-parser'
...
-  res.send('Hello world')
+  res.send('Welcome!')
+})
+
+app.get('/books', async ( req, res) => {
+  const allBooks = await Book.find()
+  res.json(allBooks)
+})
+
+app.get('/books/book/:bookID', async (req, res) => {
```
HenrikeW, line 51: I think you could make the path shorter and still have it RESTful, like
`/books/:bookID`

server.js
```diff
@@ -3,36 +3,71 @@ import bodyParser from 'body-parser'
...
+  const oneBook = await Book.findOne( { bookID: req.params.bookID })
+  if (oneBook) {
+    res.json(oneBook)
+  } else {
+    res.status(404).json({ error: 'No book found with that ID' })
+  }
+})
+
+app.get('/books/authors/:authorName', async (req, res) => {
```
HenrikeW, line 60: Thinking about the REST guidelines, I think it would make more sense to use query parameters if you look for books of a specific author. To me, your path is a bit confusing because we're kind of looking through two collections (/books and /authors) at the same time. But it's rather a gut feeling - maybe a code coach can clarify? 😁
I would rather expect for example:
/books/:bookID/authors --> to get the authors of a specific book (if it has several authors)
/authors/:authorname --> if your API didn't have books but authors to get a single author

server.js
```diff
@@ -3,36 +3,71 @@ import bodyParser from 'body-parser'
...
+  const AuthorName = req.params.authorName
+  const authorBooks = await Book.find({ authors: { $regex : new RegExp(AuthorName, "i") } })
```
HenrikeW, line 62: so cool to solve this with regex! 💯

server.js
```diff
@@ -3,36 +3,71 @@ import bodyParser from 'body-parser'
...
+
+  if ( authorBooks.length === 0) {
+    res.status(404).json('no books found by that author')
```
HenrikeW, line 65: inside the json rather make it an error object like in line 56

HenrikeW (changes requested):
You did a great job Moa and created some nice endpoints for your API! Your code is clean and easy to follow and I think the solution using Regex for the author was really cool! 

When it comes to two of your endpoints, I'm not really sure if they follow the rules for a RESTful API. I think one of them could better have been solved with a query parameter but as that was not part of the blue requirements, it might be okay anyway. As I'm not sure enough myself about the REST guidelines, I'll leave this decision to one of our code coaches and (rather as a formality) request changes. 

Other than that: Great job! Keep up the good work 💪

server.js
KarolCislowski, line 51: ID is a default key for a document, so proper endpoint would be `/books/:id`

KarolCislowski (approved):
Great Work 👍🏻

PR #82, opened 2020-12-16
project-mongo-api Anna Hasselström

annahasselstrom:
This week was fun too!  And also very good to be able to continue on from last week's project. I'm not stretching this week either. Instead I'll go back to React again and this time go over props  😀

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,11 +4,13 @@ This project follows on from my project express-api where I used a set of data l
 
 ## View it live  👀
 
+https://project-mongodb.herokuapp.com/
+
 ## The problem  🤓
 
 The API I have worked with has: 
 
-- Two different routes. 
+- Two different routes making use of both queries and params.
 - One endpoint to return a **collection** of results (array of elements)
 - One endpoint to return a **single** result (single element).
 - The API make use of Mongoose models to model the data and use these models to fetch data from the database.
--- a/README.md
+++ b/README.md
@@ -1,13 +1,24 @@
-# Mongo API Project
+# Mongo API Project  🖥
 
-Replace this readme with your own information about your project. 
+This project follows on from my project express-api where I used a set of data locally. This time I model my database using Mongoose models, and persist the data in a database.
 
-Start by briefly describing the assignment in a sentence or two. Keep it short and to the point.
+## View it live  👀
 
-## The problem
+## The problem  🤓
 
-Describe how you approached to problem, and what tools and techniques you used to solve it. How did you plan? What technologies did you use? If you had more time, what would be next?
+The API I have worked with has: 
 
-## View it live
+- Two different routes. 
+- One endpoint to return a **collection** of results (array of elements)
+- One endpoint to return a **single** result (single element).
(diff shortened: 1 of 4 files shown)


server.js
```diff
@@ -3,35 +3,85 @@ import bodyParser from 'body-parser'
...
+	const seedDatabase = async () => {
+    await Nomination.deleteMany();
+		goldenGlobesData.forEach(item => {
+      const newNomination = new Nomination(item);
+      newNomination.save();
+		})
+  }
+  seedDatabase();
+}
+
```
KarolCislowski, line 38: What happened with this indentation? 😄

server.js
```diff
@@ -3,35 +3,85 @@ import bodyParser from 'body-parser'
...
+    //const queryParameters = req.query;
+    const allNominations = await Nomination.find(req.query);
+    if (allNominations) {
+      res.json(allNominations)
+    } else {
+      res.status(404).json({ error: 'Nominations not found'})
+    }
+})
+
+app.get('/nominations/:nominee', async (req, res) => {
```
KarolCislowski, line 74: `nominee` is not a good key for this endpoint, unfortunately. When you use it like this you returning a bad set of data because you cutting off a big portion of it. For example "Avatar" has a lot of nominations, but you are showing only the first one. 

README.md
```diff
@@ -1,13 +1,26 @@
...
+- The API make use of Mongoose models to model the data and use these models to fetch data from the database.
+- The API is [RESTful](https://www.smashingmagazine.com/2018/01/understanding-using-rest-api/)
+
+## Learnings  🌟
+
+During this project I have learnt:
+
+- How to model data in Mongoose
+- How to fetch items from a Mongo database using Mongoose
+- How to seed large amounts of data to a database
```
KarolCislowski, line 25: Add the list of your endpoint, so users know how to use it ;)

KarolCislowski (changes requested):
You are very close! Small changes and it will be a very nice project 😺

README.md
annahasselstrom: Hey Karol, thank you for your review! Not sure what you mean here - can you give an example please? Thanks!

server.js
annahasselstrom: This I will fix right now! Weird😀

server.js
annahasselstrom: I understand but how do I do it when the data doesn't contain an id field?

README.md
HIPPIEKICK: It's nice with some documentation so the users know what endpoints you have. Not a requirement but a real nice-to-have!

server.js
HIPPIEKICK: If you want to use /:nominee you can do it, but you shouldn't do it together with a method that only returns a single result (the first match, even though there are many matches). As it is a requirement for this project to have an endpoint that returns a single result, you could keep this endpoint and change it to display all nominations with the help of another method but you also have to create an endpoint that returns a single result where it actually _is_ only one result e.g. a specific category, a specific year where win=true.

PR #85, opened 2020-12-17
Claire Caudwell MongoDB project

ClaireCaudwell:
**Link to my deployed API:** https://clairebookapi.herokuapp.com/documentation

**Reflections in this week:** Interesting week learning about setting up a database and accessing the data stored there with help of Mongoose when creating our API's. Not as hard as I thought it would be, but I guess we're only really scratching the surface. 

Things I want clarified: At the moment not really anything. Would like to practice working with more than one data set as I said last week also, so when I have time I will try and implement this.

### DIFF
--- a/server.js
+++ b/server.js
@@ -1,4 +1,4 @@
-import express, { query } from 'express';
+import express from 'express';
 import bodyParser from 'body-parser';
 import cors from 'cors';
 import mongoose from 'mongoose';
@@ -11,8 +11,7 @@ const mongoUrl = process.env.MONGO_URL || "mongodb://localhost/books";
 mongoose.connect(mongoUrl, { useNewUrlParser: true, useUnifiedTopology: true });
 mongoose.Promise = Promise;
 
-// Creating the model/structure of the database and what data will be stored there for each array element. This model is based on the books.json data set, but it can be based on any kind of data set. Can be string, number and boolean written as values, but which one it is depends on what the properties of the keys are written as in the initial data set.
-// Changed bookID in the data set to _id so then I could query using findById() to find a specific id of a book
+// Creating the model/structure of each of the instances that will be entered into the database. This model is based on the object structure/keys/values defined for the array elements in the books.json data set. We write the key name and the value type of each of the object keys as defined in the books.json.
 const Book = mongoose.model('Book', {
   bookID: Number,
   title: String,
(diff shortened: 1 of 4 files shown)


server.js
```diff
@@ -1,38 +1,140 @@
...
+  authors: String,
+  average_rating: Number,
+  isbn: String,
+  isbn13: String,
+  language_code: String,
+  num_pages: Number,
+  ratings_count: Number,
+  text_reviews_count: Number,
+});
+
```
annahasselstrom, line 27: I understand that you had to define the isbn as a String. But I don't understand why this is. It looks like a number to me :). Maybe the reviewing code coach can explains this?

server.js
```diff
@@ -1,38 +1,140 @@
...
 
 // Add middlewares to enable cors and json body parsing
-app.use(cors())
-app.use(bodyParser.json())
+app.use(cors());
+app.use(bodyParser.json());
+
+// First endpoint that will return the whole array of books data
+// Async and await are being used as find is asyncronous and might take some time to return the data being fetched from the server.
+app.get('/books', async (request, response) => {
```
annahasselstrom, line 56: Is find() as a method asynchronous that needs await/async as you say in your comment, or is a http request asynchronous? And we handle that with await/async?

server.js
```diff
@@ -1,38 +1,140 @@
...
+  const {author, title, language, averagerating } = request.query;
+
```
annahasselstrom, line 58: Nice! 🙌

server.js
```diff
@@ -1,38 +1,140 @@
...
+  // RegExp is a regular expression object that is used to match the author string/text added to the query parameter with a pattern. 
+  if(author) {
+    const authorRegexp = new RegExp(author, "i");
+    const queryAuthor = await Book.find({authors: authorRegexp});
+    if(queryAuthor.length === 0){
+      return response.status(404).json({ error: "Sorry that we don't have any data for that author"});
+    }
+    return response.json(queryAuthor);    
+  }
+
```
annahasselstrom, line 68: I really like that you use regex!

server.js
```diff
@@ -1,38 +1,140 @@
...
+    }
+    return response.json(queryAveRat);    
+  }
 
-// Start defining your routes here
-app.get('/', (req, res) => {
-  res.send('Hello world')
+  const books = await Book.find();
+  response.json(books);
 })
```
annahasselstrom, line 98: I see you use nested if's instead of else. Why is that better? Code coach: is there a convention here when to use what?

server.js
```diff
@@ -1,38 +1,140 @@
...
+    "https://clairebookapi.herokuapp.com/books?language=choose":  "Use this endpoint to return books written in a specific language using the title query string parameter. It will return an array with one or more elements. The following languages are valid: eng, en-GB, en-US, spa, fre, ger, ara, por, grc, mul. If the language isn't valid then you'll get an error message.",
+    
+    "https://clairebookapi.herokuapp.com/books?averagerating=choose": "Use this endpoint to return books with a specific average rating using the averagerating query string parameter. It will return an array with one or more elements. If the average rating isn't found you'll get an error message.",
+  },
+  
+  "Endpoint 2": {
+    "https://books-deployment.herokuapp.com/books/enterbookidnumber": " Use this endpoint to return books with a specific id and replace :id with a number.",
+  },
+};
+
```
annahasselstrom, line 129: I love that you write a documentation like this - so clear and helpful! 🌟

annahasselstrom (approved):
Very nicely structured! Well done Claire 🌟. I also got really inspired to learn more about regex - I see how powerful it can be. It was also interesting to see how you use nested if's and no elses - even if I don't know why we would do that 😊. We talked about the isbn and how you had to make it into a string to make it work but neither of us know why this is. Can the reviewing code coach please explain? You included documentation - great 🙌

server.js
KarolCislowski: ISBN is the international identification number and it is safer to store it as a string because it has a certain length and when someone makes mistake like this:
ISBN(as a number) + 1 = ISBN of different book.
ISBN(as a string) + 1 = ISBN1 -> error: string is too long

server.js
```diff
@@ -1,38 +1,140 @@
...
+      return response.status(404).json({ error: "Sorry that we don't have any data for that language"});
+    }
+    return response.json(queryLanguage);    
+  }
+
+  if(averagerating) {
+    const queryAveRat = await Book.find({average_rating: averagerating});
+    if(queryAveRat.length === 0){
+      return response.status(404).json({ error: "Sorry that we don't have any data for that average rating"});
+    }
```
KarolCislowski, line 91: When you use queries to filter your data you should allow your user to filter through all cases at once.

KarolCislowski (approved):
Great Work 🥇

PR #89, opened 2020-12-18
Project MongoDB

LillanEriksson:
https://lillemorsfirstmongodb.herokuapp.com/nominations

This was fun. Wish I had more time for this, but the weeks before christmas my other work takes up too much time right now.
I hope I reached blue level at least. If else, I will redo this after new years when I have more time and my brain is well-rested.

### DIFF
--- a/server.js
+++ b/server.js
@@ -1,38 +1,78 @@
-import express from 'express'
-import bodyParser from 'body-parser'
-import cors from 'cors'
-import mongoose from 'mongoose'
+import express from 'express';
+import bodyParser from 'body-parser';
+import cors from 'cors';
+import mongoose from 'mongoose';
 
 // If you're using one of our datasets, uncomment the appropriate import below
 // to get started!
-// 
-// import goldenGlobesData from './data/golden-globes.json'
+//
+import goldenGlobesData from './data/golden-globes.json';
 // import avocadoSalesData from './data/avocado-sales.json'
 // import booksData from './data/books.json'
 // import netflixData from './data/netflix-titles.json'
 // import topMusicData from './data/top-music.json'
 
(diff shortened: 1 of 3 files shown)


server.js
```diff
@@ -1,38 +1,78 @@
...
+
+//route to a single item
+
+app.get('/nominations/:nominee', async (req, res) => {
+	const singleNominee = await Nomination.findOne({
+		nominee: req.params.nominee,
+	});
+
+	res.json(singleNominee);
+});
```
KarolCislowski, line 66: `nominee` is not a good key for extracting single document, unfortunately. When you use it like this you returning a bad set of data because you cutting off a big portion of it. For example "Avatar" has a lot of nominations, but you are showing only the first one.

KarolCislowski (changes requested):
Fix this one little issue and it will be a very nice project 💪🏻

server.js
LillanEriksson: This set of data did not have an id, so what should I do?

KarolCislowski: Mongoose generates I'd for you. You can access it as `_id`

LillanEriksson: Fixed it!

server.js
```diff
@@ -1,38 +1,78 @@
...
-  res.send('Hello world')
-})
+app.get('/nominations', async (req, res) => {
+	const allNominations = await Nomination.find();
+	res.json(allNominations);
+});
+
+//route to a single item
+
+app.get('/nominations/nominee(id', async (req, res) => {
```
KarolCislowski, line 60: `'/nominations/nominee(id'` should be `/nominations/:id` you have a typo :) fix it please

KarolCislowski (approved):

PR #90, opened 2020-12-18
project-mongo-api-Therese Linden

thereselinden:
**Link:** 
https://thereselinden-mongo-book-api.herokuapp.com/


README.md
```diff
@@ -1,13 +1,35 @@
 # Mongo API Project
```
ylvan75, line 1: Clean and nice README.
- Easy to read

server.js
```diff
@@ -1,38 +1,195 @@
...
+import cors from 'cors';
+import mongoose from 'mongoose';
+import booksData from './data/books.json';
+import authorsData from './data/authors.json';
+
+const mongoUrl = process.env.MONGO_URL || "mongodb://localhost/project-mongo";
+mongoose.connect(mongoUrl, { useNewUrlParser: true, useUnifiedTopology: true });
+mongoose.Promise = Promise;
+
+//_____________create modules
```
ylvan75, line 12: I really like these comments, so clean.

server.js
```diff
@@ -1,38 +1,195 @@
...
-app.use(cors())
-app.use(bodyParser.json())
-
-// Start defining your routes here
+import express from 'express';
+import bodyParser from 'body-parser';
+import cors from 'cors';
+import mongoose from 'mongoose';
+import booksData from './data/books.json';
+import authorsData from './data/authors.json';
```
ylvan75, line 6: I think we should be proud of that we did our own data file with uniques authors.

server.js
```diff
@@ -1,38 +1,195 @@
...
+mongoose.connect(mongoUrl, { useNewUrlParser: true, useUnifiedTopology: true });
+mongoose.Promise = Promise;
+
+//_____________create modules
+const Book = mongoose.model('Book', {
+  bookID: Number,
+  title: String,
+  authors: {
+    type: mongoose.Schema.Types.ObjectId,
+    ref: 'Author',
```
ylvan75, line 18: Here as well, we should be proud of combine two collections

server.js
```diff
@@ -1,38 +1,195 @@
...
+    .send({ error: "No books found"});
+  };
+});
+
+//_____________Single book
+app.get('/books/:id', async (req, res) => {
+  const { id } = req.params;
+
+  try {
+  const singleBook = await Book.findOne({ bookID: +id });
```
ylvan75, line 121: Seems good to make sure the id is a number by doing `+id`

server.js
```diff
@@ -1,38 +1,195 @@
...
+  const pageNumber = +page || 1; 
+  const pageSize = 20;
+  const skip = pageSize * (pageNumber -1);
+
+  const authorsArray = await Author.find({
+    authors: new RegExp(author, 'i'),
+  }) 
+  .sort({ authors: 1 })
+  .limit(pageSize)
+  .skip(skip);;
```
ylvan75, line 146: Nice that you re-use the same structure as for the endpoint /books
- Sorting and pagination

server.js
```diff
@@ -1,38 +1,195 @@
...
+  res.send({ 
+    currentPage: pageNumber,
+    numberOfResults: authorsArray.length, 
+    results: authorsArray 
+  });
+}); 
+
+//_____________Single author based in ID
+app.get('/authors/:id', async (req, res) => {
+  const singleAuthor = await Author.findById(req.params.id);
```
ylvan75, line 162: Just a minor thing to be consistent. Much easier to see these things on others code then being good at it myself.

```suggestion
const { id } = req.params;
const singleAuthor = await Author.findById({ _id: id });
```

server.js
```diff
@@ -1,38 +1,195 @@
...
+  } else {
+    res
+    .status(404)
+    .json({ error: 'Author not found' });
+  }
+});
+
+//_____________Books created by author based on authors ID
+app.get('/authors/:id/books', async (req, res) => {
+  const author = await Author.findById(req.params.id);
```
ylvan75, line 174:
```suggestion
  const { id } = req.params;
const singleAuthor = await Author.findById({ _id: id });
```

server.js
```diff
@@ -1,38 +1,195 @@
...
+  if (author) {
+    const books = await Book.find({
+      authors: mongoose.Types.ObjectId(author.id)
+    })
+    .populate('authors')
+    .sort({ average_rating: -1 });
```
ylvan75, line 180: I like this!

server.js
```diff
@@ -1,38 +1,195 @@
...
+    res.json({ 
+      numberOfBooks: books.length, 
+      results: books 
+    });
+  } else {
+    res
+    .status(404)
+    .json({ error: 'authors/:id/books' });
```
ylvan75, line 188: Seems like you forgot to change the error message.

Perhaps you want something like this.

```suggestion
    .json({ error: `No books found with author id ${id}.` });
```

ylvan75 (approved):
Hi Therese, let's see if I can give you some good coding feedback. Since you and I worked together during this project and therefore of course we have chosen the same code solution for our endpoints I think it will be hard.

I start by saying that I like both your README and server.js, they are easy to read. I really like your way of doing comments.

**Things that I think we should be proud off that we managed to do/figure out:**

- Combine two collections

- Created our own data file with uniques authors - authors.json
--> Here we used books.json and with code (unfortunately I don't remember exactly how we did) we were able to filter out unique authors. That each author was an object in an array.

- How to get Heroku that runs our server.js to populate our datebase on MongoDB Cloud Atlas

It was really tough to do the code review and came up with good code improvements. I hope one of the code coaches will give us more in depth code improvements later on to reflect on and improve our backend skills.

**Two things for us to keep in mind for future projects:**

- Use const's instead of text in error handling

- To better understand when to use json or send, I saw that we both did json for the error handling for our last endpoint --> /authors/:id/books'

As always fun working with you Therese!

server.js
```diff
@@ -1,38 +1,195 @@
...
+mongoose.Promise = Promise;
+
+//_____________create modules
+const Book = mongoose.model('Book', {
+  bookID: Number,
+  title: String,
+  authors: {
+    type: mongoose.Schema.Types.ObjectId,
+    ref: 'Author',
+  },
```
KarolCislowski, line 19: 🥇 relation to another model 😻

server.js
```diff
@@ -1,38 +1,195 @@
...
+    .send({ error: `${id} is not a valid bookID`});
+  };
+});
+
+//_____________Array of authors
+app.get("/authors", async (req, res) => {
+  const { author, page } = req.query;
+  const pageNumber = +page || 1; 
+  const pageSize = 20;
+  const skip = pageSize * (pageNumber -1);
```
KarolCislowski, line 139: You used this logic again by copying the code. You should create a separate function to perform this logic and call it when needed.

KarolCislowski (approved):
Great Project! 🥇

PR #93, opened 2020-12-18
Project mongo API - Ingela

IngelaL:
It was a fun project once I had solved all my technical issues :) 

Would be great to have some more information in the project brief about MongoDB Cloud Atlas.

Check out my read me file to see all the links to the endpoints: https://github.com/IngelaL/project-mongo-api

### DIFF
--- a/server.js
+++ b/server.js
@@ -6,7 +6,7 @@ import mongoose from 'mongoose'
 // If you're using one of our datasets, uncomment the appropriate import below
 // to get started!
 // 
-// import goldenGlobesData from './data/golden-globes.json'
+import goldenGlobesData from './data/golden-globes.json'
 // import avocadoSalesData from './data/avocado-sales.json'
 // import booksData from './data/books.json'
 // import netflixData from './data/netflix-titles.json'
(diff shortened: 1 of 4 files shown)


jespa374 (changes requested):
Well done Ingela! 🥇 You seem to understand Mongoose well and it's nice to see you have played around with different endpoints and thought about the end user. Very well commented code and code which is easy to follow. I would like you to fix the indentation before approval but otherwise great job!

IngelaL:
Thanks Jessica! I have now pushed my changes.

server.js
```diff
@@ -27,9 +29,69 @@ const app = express()
...
+app.get('/nominations/nominee/:nominee', async (req, res) => {
+  const nominee = req.params.nominee
+  const singleNominee = await Data.findOne({nominee: { $regex: nominee, $options: 'i'}})
+
+  if (!singleNominee) {
+    res.status(404).json(ERROR)
+  } else {
+    res.json(singleNominee)
+  }
+})
```
KarolCislowski, line 80: `nominee` is not a good key for extracting single document, unfortunately. When you use it like this you returning a bad set of data because you cutting off a big portion of it. For example "Avatar" has a lot of nominations, but you are showing only the first one.

KarolCislowski (changes requested):
Please add a good endpoint which will return a single document.

server.js
IngelaL: Since there aren't any id:s for this dataset I deleted the endpoint and added another one on line 70: /nominations/:year/:category/winner.

KarolCislowski: Yes, you have it. Mongoose creates it automatically and names it `_id` you can see it here: https://project-mongo-api-ingela.herokuapp.com/nominations 
The easiest way to connect to it would be to use https://mongoosejs.com/docs/api.html#model_Model.findById

IngelaL: Thanks! Think I got it now. Just pushed new version with endpoint /nominations/nominee/:id

KarolCislowski (approved):
Great Job 😸

PR #94, opened 2020-12-18
Pull req for mongodb project

Katsiez:


server.js
```diff
@@ -1,38 +1,115 @@
...
+    response.status(503).json({ error: "Service unavailable" });
+  }
+});
+
+app.get('/books', async (request, response) => {
+  const allBooks = await Book.find();
+  response.json(allBooks);
+});
+
+app.get("/books/find/:bookID", async (request, response) => {
```
moablomkvist, line 64: think about whether it should say something other than "find" in your endpoint. maybe "id" ?, to know which number it is then just by reading.

server.js
```diff
@@ -1,38 +1,115 @@
...
+app.get("/books/isbn/:isbn", async (request, response) => {
+  try {
+    const { isbn } = request.params;
+    const book = await Book.findOne({ isbn: isbn });
+    if (book) {
+      response.json(book);
+    } else {
+      response
+        .status(404)
+        .json({ error: `No books found with the following isbn=${isbn}` });
```
moablomkvist, line 91: I like that the error message refers to "the following isbn ... " makes it even more helpful. clever!

server.js
```diff
@@ -1,38 +1,115 @@
...
+    }
+  } catch (err) {
+    response.status(400).json({ error: "invalid request" });
+  }
+});
 
+app.get("/books/authors/:author", async (request, response) => {
+  const searchedAuthor = request.params.author;
+  const thisBookAuthor = await Book.find({
+    authors: { $regex: new RegExp(searchedAuthor, "i") },
```
moablomkvist, line 101: yay! Regex!

server.js
```diff
@@ -1,38 +1,115 @@
...
+  });
+  if (thisBookAuthor.length === 0) {
+    response
+      .status(404)
+      .json(
+        "No books have been found by such author. Perhaps try searching another author?"
+      );
+  }
+  response.json(thisBookAuthor);
+});
```
moablomkvist, line 111: Great work Kat! and Happy new year!

server.js
KarolCislowski, line 64: `id` is a key for this model so just use `/books/:id` for another field you should ad name of the field like this `/books/isbn/:isbn`

KarolCislowski (approved):
Great Work 👍🏻

PR #95, opened 2020-12-18
books-mongoDb-Tara

trangtara:


server.js
```diff
@@ -3,36 +3,260 @@ import bodyParser from 'body-parser'
...
 app.get('/', (req, res) => {
-  res.send('Hello world')
+    
+  
+  if(!res) {
+    res
+    .status(404)
+    .send({ error: 'Oops! Something goes wrong. Try again later!'})
+  }
+  res.send(myEndPoints(app))
```
the-j-curl, line 79: This is a nice feature to show all available endpoints on the root page

server.js
```diff
@@ -3,36 +3,260 @@ import bodyParser from 'body-parser'
...
+})
+//API endpoints for books with query options for author, title; otherwise, display 20 data per page (sorted by average_rating) and allow user to request for more pages
+app.get('/books', async (req, res) => {
+  const { title, author } = req.query
+
+  const queryTitleRegex = new RegExp(title, 'i')
+  const queryAuthorRegex = new RegExp(author, 'i')
```
the-j-curl, line 86: What do these two lines of code do? Don't think I have seen this before

server.js
```diff
@@ -3,36 +3,260 @@ import bodyParser from 'body-parser'
...
+      results: []
+    }
+
+    const books = await Book.find()
+      .sort({ average_rating: -1 })
+      .skip(skipData)
+      .limit(pageSize)
+    
+    if (!books.length) {
+      res.json({ error: 'Oops! Page not found'})
```
the-j-curl, line 135: You should send a status code with this error message. To provide more info to the client, and also to be consistent with your other error messages

server.js
```diff
@@ -3,36 +3,260 @@ import bodyParser from 'body-parser'
...
+  if(!res) {
+    res
+    .status(404)
+    .send({ error: 'Oops! Something goes wrong. Try again later!'})
+  }
+  res.send(myEndPoints(app))
+})
+//API endpoints for books with query options for author, title; otherwise, display 20 data per page (sorted by average_rating) and allow user to request for more pages
+app.get('/books', async (req, res) => {
+  const { title, author } = req.query
```
the-j-curl, line 83: One tip: you can use the following line of code.  This is a clean way of setting queries. This allows all database fields to be queries
`const allBooks = await Book.find(req.query)`

server.js
```diff
@@ -3,36 +3,260 @@ import bodyParser from 'body-parser'
...
+    }
+  } catch (err) {
+    res.status(400).json({ error: `${isbn} is an invalid isbn`})
+  }
+})
+
+//APIs BELOW IS TO PRACTISE MONGODB AGGREGATE FUNCTIONS
+
+//sort and group books by language. If there is no lang query, group the data by language, authors and titles
+app.get('/books/language/search', async (req, res) => {
```
the-j-curl, line 176: Cool that you can search by language

server.js
```diff
@@ -3,36 +3,260 @@ import bodyParser from 'body-parser'
...
+    }
+  } catch (err) {
+    res.status(400).json(ERROR)
+  }
+})
+//CREATE NEW COLLECTION OF TOP 40 ENGLISH BOOK
+//create new data collection called top40EngBook from the current Book collection, then create an API endpoints to printout top 40 English book from new collection
+app.get('/top40rating/english', async (req, res) => {
+  try {
+    const top40EngBooks = await Book.aggregate([
```
the-j-curl, line 215: This is a nice endpoint too

the-j-curl (approved):
Hi Tara.

Really nice work with this project, I don't have many comments to add. Your code looks tidy and you have created a really in depth API with lots of endpoints.

One thing I noticed is that you have a lot of code when defining endpoints, which looks complex and might be able to be simplified. But I don't understand backend well enough yet to make a certain suggestions. I did point out one good tip when defining query param's. 
Maybe a code coach could make some suggestions to simplify your code?

KarolCislowski (approved):
Great Work 🥇

PR #97, opened 2020-12-18
project-mongo-api Ellen

ellenbjo:
This has been a fun week! I decided to try out the pagination and found a medium article on it that was helpful. Still missing some error handling for the query-params in the /books endpoint, will come back to that later.

### DIFF
--- a/models/book.js
+++ b/models/book.js
@@ -2,13 +2,13 @@ import mongoose from 'mongoose'
 //second param --> object--> called schema
 //start with capital letter reserved for Models in Mongoose.
 //string is going to be name of collection with the 1st string param -->  start with lower case and with and ends with an s
-//what to do with isbn? one of the isbn:s is not a number. it has an X
+//isbn is a string because some items have letters 
 const Book = new mongoose.model('Book', {
   bookID: Number,
   title: String,
   authors: String,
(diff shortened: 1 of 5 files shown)


models/book.js
```diff
@@ -0,0 +1,19 @@
...
+  average_rating: Number,
+  isbn: String,
+  isbn13: Number,
+  language_code: String,
+  num_pages: Number,
+  ratings_count: Number,
+  text_reviews_count: Number
+})
+
+export default Book
```
KarolCislowski, line 19: External Models makes it super clean!

server.js
```diff
@@ -3,33 +3,99 @@ import bodyParser from 'body-parser'
...
+      totalPages: Math.ceil(count / limit),
+      currentPage: page
+    })
+  } catch(err) {
+    res.status(404).json({ error: 'books not found'})
+  }
+})
+
+//findOne returns one element. returns an object instead of an array with an object
+app.get('/books/book/:id', async (req, res) => {
```
KarolCislowski, line 88: `id` is a default key for this document so better endpoint URL would be `/books/:id`

KarolCislowski (approved):
Great Work 💯

PR #102, opened 2020-12-18
Caroline Birgersson - Mongo project

carolinebirgersson:


liza-minican (approved):
Hey Caroline, you did a great job!!! I enjoyed trying out your endpoints! There is not much i can suggest, maybe comment your code a bit more extensively..you could explain what your regex does to allow finding an author by one letter or your errors..which one will be shown in which case.  Other than that keep up good work!!

PR #109, opened 2020-12-20
Mongo API Books-project database

Snabbina:
More info and link se README.


server.js
```diff
@@ -1,38 +1,70 @@
...
+//Author model and connection to database
+const mongoUrl = process.env.MONGO_URL || "mongodb://localhost/project-books";
+mongoose.connect(mongoUrl, { useNewUrlParser: true, useUnifiedTopology: true });
+mongoose.Promise = Promise;
+
+const port = process.env.PORT || 8080;
+const app = express();
+
+//A model
+const Author = mongoose.model("Author", {
```
VanessaSue27, line 17: Since this is a database containing different book objects, I think it would make more sense to name the main model here Book instead of Author 📚
It can be confusing, since with this name I am expecting to get data on a bunch of different authors, but for this data set this is not the case.

Then in that case, if you have a main Book model instead, you could take advantage of all the information that is provided to us in the dataset, and add more properties to this model, like: "average_rating", "num_pages", "ratings_count", etc. 🤓

server.js
```diff
@@ -1,38 +1,70 @@
...
+  const seedDatabase = async () => {
+    await Author.deleteMany();
+
+    booksData.forEach((item) => {
+      const newAuthor = new Author(item);
+      newAuthor.save();
+    });
+  };
+  seedDatabase();
+}
```
VanessaSue27, line 37: Missing ; I believe 👀

server.js
```diff
@@ -1,38 +1,70 @@
...
+
+// Routes starts here
+app.get("/", (req, res) => {
+  res.send("Welcome, this is the database full of great authors and books");
+});
+
+//Get all authors
+app.get("/authors", async (req, res) => {
+  const allAuthors = await Author.find(req.query);
```
VanessaSue27, line 46: If the main purpose of this endpoint is to show all authors, then I think you don't need to have req.query in between the parenthesis. find() alone will do the trick.
Unless you have something else in mind, since having req.query in there makes all properties per entry available from the request URL? 🧐

server.js
```diff
@@ -1,38 +1,70 @@
...
+  console.log(allAuthors);
```
VanessaSue27, line 47: console.logs can be removed before submitting final project

server.js
```diff
@@ -1,38 +1,70 @@
...
+
+  if (allAuthors.length > 0) {
+    res.json(allAuthors);
+  } else {
+    res.status(404).json({ error: "Sorry, could not find the data" });
```
VanessaSue27, line 52: Nice implementing error responses! ✨

server.js
```diff
@@ -1,38 +1,70 @@
...
+  }
+});
+
+//Get a single title for an author
+app.get("/authors/title/:title", async (req, res) => {
+  const singleTitle = await Author.findOne({ title: req.params.title });
```
VanessaSue27, line 58: I only have one observation for this endpoint: since I know in theory this endpoint works and the code is correct. 💡

The problem I see is that, since you have declared "title: req.params.title", in order for the API to find a match, the title in the request URL has to be typed exactly the same way that the title appears in the database and this is very tricky.

For example, I did a search for "Harry Potter" and I got the message that no titles could be found, even though I know there are several Harry Potter books in this database.

In this scenario, I would suggest implementing a regular expression that makes that search case-insensitive, and finds matches even if the string is partially included in the original title. 💥
That might give us more desirable results.

VanessaSue27 (approved):
Hello Sabina!

You've achieved a very clean and straight to the point project 💪
I have left some comments on the things that I noticed, some suggestions that could make this API even more functional 💥

Good job getting the whole setup to work both with Heroku and MongoDB Cloud Atlas ✨
Since all basic requirements are achieved, I'm approving this project ⭐

// Vanessa 🍉

server.js
Snabbina: Cool thank you for the tips!

server.js
Snabbina: fixed!

server.js
Snabbina: fixed.

server.js
Snabbina: Yes, regex might to the trick!

server.js
KarolCislowski: @VanessaSue27 is right. Authors collection should contain the list of authors, not books. When you are designing your data you need to ask yourself a question "what I will store here".

You can create separate Authors collection for example like this:
{ id: String, name: String, books: []}

server.js
```diff
@@ -1,38 +1,69 @@
...
+  seedDatabase();
+};
+
+// Routes starts here
+app.get("/", (req, res) => {
+  res.send("Welcome, this is the database full of great authors and books");
+});
+
+//Get all authors
+app.get("/authors", async (req, res) => {
```
KarolCislowski, line 45: When you change your data, you will need to change your endpoints as well.

KarolCislowski (changes requested):
Please change your naming convention and then it will be Great project 😸

Snabbina:
@KarolCislowski, thank you for your review I have made the changes and tested it out in Postman, please let me know if its is ok now =) /S

KarolCislowski (approved):
Now it is Great 🥇

PR #111, opened 2020-12-20
Week 18 - Project mongo-api - Karolin Albinsson

karolinalbinsson:


server.js
```diff
@@ -1,38 +1,218 @@
...
+const port = process.env.PORT || 8080;
+const app = express();
 
 // Add middlewares to enable cors and json body parsing
-app.use(cors())
-app.use(bodyParser.json())
+app.use(cors());
+app.use(bodyParser.json());
+const myEndpoints = require("express-list-endpoints");
+
```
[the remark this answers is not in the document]
thereselinden:
```suggestion
//_____________Error message in case server is down
app.use((req, res, next) => {
  if(mongoose.connection.readyState === 1) {
    next();
  } else {
    res
    .status(503)
    .send({ error: "service unavailable"});
  };
});
```

server.js
```diff
@@ -1,38 +1,218 @@
...
+});
+
+/*COLLECTIONS*/
+//Get all the albums from the database .
+app.get("/api/albums", async (req, res) => {
+	const filterYear = req.query.year;
+	const filterArtist = req.query.artist;
+	const filterGenre = req.query.genre;
+	const filterYearTo = req.query.yearTo;
+	const filterYearFrom = req.query.yearFrom;
```
thereselinden, line 62: This could be also be written as const { year, artist, genre, yearTo, yearFrom, sortBy, sortOrder, page } = req.body

server.js
```diff
@@ -1,38 +1,218 @@
...
+//Get by placement in list
+app.get("/api/albums/placement/:placement", async (req, res) => {
+	const placement = req.params.placement;
+	const singleAlbumFiltered = await Album.findOne({ number: placement });
+	if (singleAlbumFiltered.length > 0) {
+		res.json(singleAlbumFiltered);
+	} else {
+		res.status(404).send(errorMsg);
+	}
+});
```
thereselinden, line 199: Finding it hard to understand what placement is. If I try `api/albums/placement/4` it will return an error message. 
If I try to add _id instead of 4 the server is loading for a long time, one way to prevent this is by using try catch, then the server will catch the error if not matching Model requirements.

server.js
```diff
@@ -1,38 +1,218 @@
...
 
-// Start defining your routes here
-app.get('/', (req, res) => {
-  res.send('Hello world')
-})
+//Get by album title
+app.get("/api/albums/title/:title", async (req, res) => {
```
thereselinden, line 202: Can't get this endpoint to work. If I write `/api/albums/title/Revolver ` I receive the error message.

thereselinden (approved):
Hi Karolin,

I can see that you put in a lot of work to this project. A lot of different endpoints and several queries. ⭐️  Reflection from reading your code is the importance on good documentation to be able to understand how to use the different endpoints. That is something I will take with me for future projects. 

**Some feedback:** 
- I can't get some of the endpoints to work : 
`/api/albums/placement/:placement`
`/api/albums/title/:title`

- One thing that could be nice is when I query for artist (album name) is that you use include so I can shorten "the beatles" to just "beatles". 

- Functions used for the first endpoint ` api/albums` is very large, maybe this could be written in a different way to make it shorter and more readable?

karolinalbinsson:
thank for disovering the bugs, now fixed the broken endpoints placement and title 👍

KarolCislowski (approved):
Great Job 🥇

PR #112, opened 2020-12-20
project-mongo-api - Jamie Cook

the-j-curl:
See readme for available endpoints.

My API can be found here:
https://jamie-albums-api.herokuapp.com/

### DIFF
--- a/server.js
+++ b/server.js
@@ -1,38 +1,40 @@
-import express from 'express'
-import bodyParser from 'body-parser'
-import cors from 'cors'
-import mongoose from 'mongoose'
+import express from "express";
+import bodyParser from "body-parser";
+import cors from "cors";
+import mongoose from "mongoose";
 
-// If you're using one of our datasets, uncomment the appropriate import below
-// to get started!
-// 
-// import goldenGlobesData from './data/golden-globes.json'
-// import avocadoSalesData from './data/avocado-sales.json'
-// import booksData from './data/books.json'
-// import netflixData from './data/netflix-titles.json'
-// import topMusicData from './data/top-music.json'
+import albumData from "./data/rolling-stone-top-500-albums.json";
 
-const mongoUrl = process.env.MONGO_URL || "mongodb://localhost/project-mongo"
-mongoose.connect(mongoUrl, { useNewUrlParser: true, useUnifiedTopology: true })
-mongoose.Promise = Promise
+const mongoUrl = process.env.MONGO_URL || "mongodb://localhost/project-mongo";
+mongoose.connect(mongoUrl, { useNewUrlParser: true, useUnifiedTopology: true });
+mongoose.Promise = Promise;
 
-// Defines the port the app will run on. Defaults to 8080, but can be 
-// overridden when starting the server. For example:
-//
-//   PORT=9000 npm start
-const port = process.env.PORT || 8080
-const app = express()
+const port = process.env.PORT || 8080;
+const app = express();
 
 // Add middlewares to enable cors and json body parsing
-app.use(cors())
-app.use(bodyParser.json())
+app.use(cors());
+app.use(bodyParser.json());
+
+const Album = new mongoose.model("Album", {
+  position: Number,
+  artist: String,
+  albumName: String,
+  label: String,
+  year: Number,
+  critic: String,
+});
+
+const populateDatabase = () => {
+  Album.deleteMany(); //DeleteMany is a method provided by Mongoose package
+};
 
 // Start defining your routes here
-app.get('/', (req, res) => {
-  res.send('Hello world')
-})
+app.get("/", (req, res) => {
+  res.send("Hello world");
+});
 
 // Start the server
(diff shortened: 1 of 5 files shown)


server.js
```diff
@@ -1,38 +1,89 @@
...
-
-// Start the server
+import express from "express";
+import bodyParser from "body-parser";
+import cors from "cors";
+import mongoose from "mongoose";
+
+import albumData from "./data/rolling-stone-top-500-albums.json";
+
+const mongoUrl = process.env.MONGO_URL || "mongodb://localhost/project-mongo";
```
karolinalbinsson, line 8: On this row, you could add the URL to the deployed database instead of your locahost url.

server.js
```diff
@@ -1,38 +1,89 @@
...
+      const newAlbum = new Album(item);
+      newAlbum.save();
+    });
+  };
+  seedDatabase();
+}
+
+const port = process.env.PORT || 8082;
+const app = express();
+const listEndpoints = require("express-list-endpoints");
```
karolinalbinsson, line 35: This is a nice touch including the available endpoints on "start" 👍

server.js
```diff
@@ -1,38 +1,89 @@
...
+});
+
+app.get("/", (req, res) => {
+  res.send(listEndpoints(app));
+});
+
+app.get("/albums", async (req, res) => {
+  const allAlbums = await Album.find(req.query); // Clean way of setting queries. This allows all database fields to be queries
+  if (allAlbums) {
+    res.json({
```
karolinalbinsson, line 55: Since you are providing a status where somethins goes wrong, you could also include a "200" status when the request is OK.

server.js
```diff
@@ -1,38 +1,89 @@
...
+  const singleAlbum = await Album.findOne({ position: +position });
+  if (singleAlbum) {
+    res.json(singleAlbum);
+  } else {
+    res.status(404).send({ error: `No album found for position: ${position}` });
+  }
+});
+
+app.get("/albumid/:id", async (req, res) => {
+  try {
```
karolinalbinsson, line 75: Here you are using try/catch, and in the endpoint above you are using if/else. Is there a reason? If not -maybe you could be consistent in only using one of the kinds?

server.js
```diff
@@ -1,38 +1,89 @@
...
+
+const port = process.env.PORT || 8082;
+const app = express();
+const listEndpoints = require("express-list-endpoints");
+
+app.use(cors());
+app.use(bodyParser.json());
+
+app.use((req, res, next) => {
+  if (mongoose.connection.readyState === 1) {
```
karolinalbinsson, line 41: This was a nice touch checking the db-connection :)

server.js
```diff
@@ -1,38 +1,89 @@
...
+    res.status(503).json({ error: `Service unavailable` });
+  }
+});
+
+app.get("/", (req, res) => {
+  res.send(listEndpoints(app));
+});
+
+app.get("/albums", async (req, res) => {
+  const allAlbums = await Album.find(req.query); // Clean way of setting queries. This allows all database fields to be queries
```
karolinalbinsson, line 53: This was cool!  Saves many lines of code. Gonna steal that 👍

karolinalbinsson (changes requested):
Hi Jamie, great job with the mongo-api project!
I didn't have a lot of remarks. 

+ I think your code is clean, you are being consistent with semi-colons and indentation. 
+ Your routes are restful.
+ The query-functionality is very handy!

To think about:
I think that you could be consistent in your responses to the user - you are passing a status-code when things go wrong, so perhaps do the same when the result is ok also. 

One change is needed for approval: 
data/football-results.json - remove this, since it is not used. 


//Karolin

server.js
the-j-curl: FYI I have saved my deployed database as an enviroment variable. So my live url is saved as MONGO_URL. I can just remove this from heroku and continue to use my local database if I need to do any local changes before pushing to live environment.

server.js
the-j-curl: Thanks for this tip 👍

server.js
the-j-curl: Agreed. I feel like I have a better understanding of this now

karolinalbinsson (approved):

server.js
KarolCislowski, line 53: 🥇

server.js
```diff
@@ -1,38 +1,89 @@
...
+  const { position } = req.params;
+  const singleAlbum = await Album.findOne({ position: +position });
+  if (singleAlbum) {
+    res.status(200).json(singleAlbum);
+  } else {
+    res.status(404).send({ error: `No album found for position: ${position}` });
+  }
+});
+
+app.get("/albumid/:id", async (req, res) => {
```
KarolCislowski, line 74: this endpoint is not in REST convention. It should be `/albums/:id`

KarolCislowski (approved):
Great Job 🥇

PR #114, opened 2020-12-20
Project Mongo API - Linda

lindahz:
Link to my deployed project:
https://lindas-project-mongo-api.herokuapp.com/

I had limited time this week, my focus was to achieve only blue level requirements.

Nothing I can think of.


server.js
```diff
@@ -3,35 +3,100 @@ import bodyParser from 'body-parser'
...
+  const singleID = await Character.findOne({ id: id })
+  if (singleID) {
+    res.json(singleID)
+  } else {
+    res.status(404).json(ERROR_CHARACTERS_NOT_FOUND)
+  }
+})
+
+// Shows all characters from a specific house, not case sensitive
+app.get('/characters/:house', async (req, res) => {
```
amandaflod, line 90: You could change the path to /characters/house/:house for better consistency, as you did so in the other path!

server.js
```diff
@@ -3,35 +3,100 @@ import bodyParser from 'body-parser'
...
-// import goldenGlobesData from './data/golden-globes.json'
-// import avocadoSalesData from './data/avocado-sales.json'
-// import booksData from './data/books.json'
-// import netflixData from './data/netflix-titles.json'
-// import topMusicData from './data/top-music.json'
-
-const mongoUrl = process.env.MONGO_URL || "mongodb://localhost/project-mongo"
+import harryData from './data/harry-potter-characters.json'
+
+const ERROR_CHARACTERS_NOT_FOUND = {error : 'No character results were found, please try again.'}
```
amandaflod, line 8: Good idea to create this variable!

server.js
```diff
@@ -3,35 +3,100 @@ import bodyParser from 'body-parser'
...
+    res.json(singleID)
+  } else {
+    res.status(404).json(ERROR_CHARACTERS_NOT_FOUND)
+  }
+})
+
+// Shows all characters from a specific house, not case sensitive
+app.get('/characters/:house', async (req, res) => {
+  const { house } = req.params
+  const houses = await Character.find({ house: { $regex: house, $options: "i"} })
```
amandaflod, line 92: Great use of regexp to create a simple search!

amandaflod (commented):
As always Linda - great job! Despite your time constraints that week you managed to build a well-working API and write smart and structured code! 😊 There really isn't much to comment on, I think you did good ❤️ (and sorry for late review!!!)

server.js
```diff
@@ -3,35 +3,100 @@ import bodyParser from 'body-parser'
...
+  const characters = await Character.find(req.query) 
+  if (characters.length === 0) {
+    res.status(404).json(ERROR_CHARACTERS_NOT_FOUND)
+  } else {
+    res.json(characters)
+  }
+})
+
+// Shows one character with unique id
+app.get('/characters/id/:id', async (req, res) => {
```
KarolCislowski, line 79: `id` is a default key so you should use endpoint `characters/:id`

server.js
KarolCislowski: and here `house` is not the identification key for this collection so you should use `/characters/houses/:house`

KarolCislowski (approved):
Great job here 🥇

PR #115, opened 2020-12-20
Nathalie 

nathalienordstrom:
https://mongo-api-express.herokuapp.com

### DIFF
--- a/server.js
+++ b/server.js
@@ -3,38 +3,11 @@ import bodyParser from 'body-parser'
 import cors from 'cors'
 import mongoose from 'mongoose'
 
-import netflixData from './data/netflix-titles.json'
+import membersData from './data/groData.json'
 
-const mongoUrl = process.env.MONGO_URL || "mongodb://localhost/books"
+const mongoUrl = process.env.MONGO_URL || "mongodb://localhost/gro-project"
 mongoose.connect(mongoUrl, { useNewUrlParser: true, useUnifiedTopology: true })
 mongoose.Promise = Promise
-
-const Director = mongoose.model('Director', {
-  name: String,
-})
-
-const Title = mongoose.model('Title', {
-  title: String,
-
-  director: {
-    type: mongoose.Schema.Types.ObjectId,
-    ref: 'Director'
-  }
-})
-
-if (process.env.RESET_DATABASE) {
-  const seedDatabase = async () => {
-    await Director.deleteMany()
-
-    netflixData.forEach(item => {
-      const director = new director(item)
-       director.save();
-    })
-  }
-
-  seedDatabase()
-}
-
 
 const port = process.env.PORT || 8080
 const app = express()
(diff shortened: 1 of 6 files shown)


server.js
```diff
@@ -3,26 +3,43 @@ import bodyParser from 'body-parser'
...
+})
+
+const Title = mongoose.model('Title', {
+  title: String,
+
+  director: {
+    type: mongoose.Schema.Types.ObjectId,
+    ref: 'Director'
+  }
+})
```
AxelHammarback, line 23: I think you're off to a good start, but I'm not sure this is 100% the right way to do it. From what I've understood, the "Model" (in your case the Director and the Title) should rather be a model of the data in your JSON file. 

And in the Netflix json data file, it's all listing "Shows", sort of. 

It's confusing and complicated, I know, but maybe something like this:

```
const Show = new mongoose.model('Show', {
  show_id: Number,
  title: String,
  director: String,
  cast: String,
  country: String,
  date_added: String,
  release_year: Number,
  rating: String,
  duration: String,
  listed_in: String,
  description: String,
  type: String,
})
```

server.js
```diff
@@ -3,26 +3,43 @@ import bodyParser from 'body-parser'
...
+    await Director.deleteMany()
+
+    netflixData.forEach(item => {
+      const director = new director(item)
+       director.save();
+    })
+  }
+
+  seedDatabase()
+}
```
AxelHammarback, line 36: Well done, nice-looking code for seeding the database! But if I was correct with the "Show" above (please correct me if I'm wrong, Maks/Matilda/Van/Whoever reads this), then the seeding would look something like:

```
if (process.env.RESET_DATABASE) {
  const seedDatabase = async () => {
    await Show.deleteMany()

    netflixData.forEach(item => {
      const newShow = new Show(item)
      newShow.save();
    })
  }
  seedDatabase()
}
```

server.js
```diff
@@ -32,6 +49,38 @@ app.get('/', (req, res) => {
   res.send('Hello world')
 })
 
+app.get('/directors', async (req, res) => {
+  const directors = await Director.find()
+  res.json(directors)
+})
```
AxelHammarback, line 55: I'm still on the "shows" track – so if you wanted to list all "shows" (we'll figure out how to list the directors in the next endpoint), a base endpoint could look like this:

```
// ENDPOINT 1: List all shows
// Example: http://localhost:8080/shows
app.get('/shows', async (req, res) => {
  const allShows = await Show.find(req.query)
  res.json(allShows)
})
```
So pretty much the way you did for the directors, but with `find(req.query)` instead of just `find()`. Well done! 🙂

server.js
```diff
@@ -32,6 +49,38 @@ app.get('/', (req, res) => {
...
+
+app.get('/directors/:id', async (req, res) => {
+  const director = await Director.findById(req.params.id)
+  if (director) {
+    res.json(director)
+  } else {
+    res.status(404).json({ error: 'Author not found' })
+  }
+
+})
```
AxelHammarback, line 65: Ok, and here's where the magic happens. Since a "director" is sort of a characteristic of a "show", a suitable RESTful endpoint could be: 
`localhost:8080/shows/directors/<the director's name>`

Seeing this, the endpoint to list directors could be something like: 

```
// ENDPOINT 2: List all shows of a certain director.
// **Example: http://localhost:8080/shows/directors/Abhishek%20Sharma**
app.get('/shows/directors/:director', async (req, res) => {
  const {director} = req.params
  const filteredShows = await Show.find({director: director})
  res.json(filteredShows)
})
```

AxelHammarback (commented):
Hey Nathalie, I think this is a great start! I couldn't get it working on my machine, but I think that has to do with the way the database is seeded, which looks slightly off. I added in the comments _one_ way to do it which works (but I'm sure there are many others). If you would like to take a look at that example in its entirety, I uploaded the file here: https://easyupload.io/a6ewi5

Take care 🌟

KarolCislowski (changes requested):
Your hosted app does not return any data. Please check your MongoDB configuration - probably you cant connect to it.

HIPPIEKICK:
Still no data. Have you followed this [guide](https://www.notion.so/technigo/Guide-Deploying-databases-fc4f1c0b00bf474aa8aeae2f343b4adb), and added the `MONGO_URL` and the `RESET_DATABASE=true` to heroku's config variables?

PR #119, opened 2020-12-20
Project-mongo-api-joel

JHarde:


server.js
```diff
@@ -1,38 +1,77 @@
...
+  res.send(listEndpoints(app));
+});
+
+app.get("/shows", async (req, res) => {
+  const shows = await Show.find();
+  res.json(shows);
+});
+
+app.get("/shows/:title", async (req, res) => {
+  const searchTitle = await Show.findOne({ title: req.params.title });
```
BamStugendal, line 56: Nice that you used `.findOne()`

server.js
```diff
@@ -1,38 +1,77 @@
...
+  if (searchTitle) {
+    res.json(searchTitle);
+  } else {
+    res.status(404).json({ error: "Title not found" });
```
BamStugendal, line 60: Good that you show status 👍

server.js
```diff
@@ -1,38 +1,77 @@
...
+app.get("/", (req, res) => {
+  res.send(listEndpoints(app));
+});
+
+app.get("/shows", async (req, res) => {
+  const shows = await Show.find();
+  res.json(shows);
+});
+
+app.get("/shows/:title", async (req, res) => {
```
KarolCislowski, line 55: `title` is not a good key for extracting single document, unfortunately. When you use it like this you returning a bad set of data because you can cut off a big portion of it. You can have a lot of shows with the same title, for example, remakes or title is so generic it is reused hundreds of time.

KarolCislowski (changes requested):
Please add a proper endpoint to a single document and your app will be really nice 😸

JHarde:
Makes sense, thanks for the input! Changed the endpoint! :)

KarolCislowski (approved):
Great project 💯

PR #120, opened 2020-12-20
Mongo API - Golden globe awards

destinyleia:


Snabbina (changes requested):
Hey Destiny!

So cool that you now have leveled-up your API skills and learnt to store and retrieve data from and database and use that data to produce a RESTful API!

As I understand it, you have met the requirements of:
- Your API should have at least 2 routes. And theoretically met the rest of the requirements of:
- A minimum of one endpoint to return an array of elements.
- A minimum of one endpoint to return a single element.
- Your API should make use of Mongoose models to model your data and use these models to fetch data from the database.
- Your API should be [RESTful]

(There was no link to a deployed version to test it out though!?)

So I guess in order to be approved, you need to to the parts from the general requirements too:
- Your API should be deployed to Heroku or similar hosting service.
- Your database should be deployed using mongo cloud or similar.

You have some guidelines here how to do this: https://www.notion.so/Guide-Deploying-databases-fc4f1c0b00bf474aa8aeae2f343b4adb


A tip is also to make more commits, after each "step" in your coding, instead of a single big one, then you can backtrack if something goes wrong and that you have a fresh backup of the code all the time. Also try to describe your approach more in the comments when you commit or in your README, sometimes that helps the learning and building process as well.

I know it's tough sometimes, but hang in there Destiny!  

Best of the best to you/ Sabina

server.js
```diff
@@ -27,11 +44,33 @@ const app = express()
...
+  const getNominations = await Nomination.find(req.query)
+
+  if (getNominations) {
+    res.json(getNominations)
+  } else {
+    res.status(404).json({ error: 'There is no nomination like this'})
+  }
+})
+
+app.get('/nominations/:year', async (req, res) => {
```
KarolCislowski, line 63: `year` is not a good key for extracting single document, unfortunately. 

KarolCislowski (changes requested):
Please provide a link to the hosted version of your app 😺 and add an endpoint to retrieve a single document.

server.js
HIPPIEKICK: Hi Destiny! Jumping into the conversation here ;) There's no data on the deployed API! An idea for an endpoint that returns a single result can be to have params of year and category and then show the winner?

destinyleia: Hi Matilda, 
Are you sure? I was just checking and I got back - here's one of the links https://golden-globe-api-mongo.herokuapp.com/nominees :)

HIPPIEKICK: You're right Destiny, it seems to be deployed OK! Just fix so that the endpoint shows only one result!

destinyleia: Fixed! :)

server.js
```diff
@@ -27,9 +41,47 @@ const app = express()
...
+      year_award: year,
+      category: category,
+      win: true
+    });
+  if (filteredNominees) {
+    res.json(filteredNominees)
+  } else {
+    res.status(404).json({ error: 'Not found' })
+  }
 })
```
HIPPIEKICK, line 85: The logic seems fine, but you should not hard code a query param into the path like that. Did you get this endpoint to work? You could add the win as a query param under the previous endpoint (since this is a duplicate besides the win=true).

PR #123, opened 2020-12-20
Hanna - Project Mongo API

Hanna-Skeppe:
https://hanna-mongodb-netflix-api.herokuapp.com/netflix-titles


server.js
```diff
@@ -27,9 +19,84 @@ const app = express()
...
+//Query-PARAM TO GET TITLES FROM COUNTRY: /netflix-titles?country=germany
+//EXAMPLE TO FIND TITLE(S) BY DIRECTOR & ACTOR /netflix-titles?actor=niro&director=scorsese
+app.get('/netflix-titles', async (req, res) => {
+  const { title, director, actor, type, country } = req.query;
+  const allTitles = await NetflixTitle.find({
+    title: new RegExp(title, 'i'), //regexp: 'i' is 'ignore case'
+    director: new RegExp(director, 'i'),
+    cast: new RegExp(actor, 'i'),
+    type: new RegExp(type, 'i'),
+    country: new RegExp(country, 'i')
```
idsintehittapa, line 66: Cool with the RegExp! 

You have choose some of the keys in the database and excluded others, I am not sure if this was to try it out maybe? Consider doing a map to include all the keys 😄

server.js
```diff
@@ -27,9 +19,84 @@ const app = express()
...
+    .sort({ release_year: -1 }); //sort desc.
+  if (allTitles.length === 0) {
+    res.status(404).json({ error: 'Could not find any Netflix-titles to show' })
+  } else {
+    res.json(allTitles);
+  }
+})
+
+//YEAR-ENDPOINT: get all titles from a specific year.
+app.get('/netflix-titles/year/:year', async (req, res) => {
```
idsintehittapa, line 77: I think this could have been a query too for consistency maybe? 
And if you really wanted this to be an endpoint consider making it into 
`app.get('/netflix-titles/:year', async (req, res)`
To make it a bit more easy for the user

server.js
```diff
@@ -27,9 +19,84 @@ const app = express()
...
+  const titlesByYear = await NetflixTitle.find({
+    release_year: year
+  })
+
+  if (titlesByYear.length === 0) {
+    res.status(404).json({ error: `Could not find any Netflix-title from year: ${year}` })
+  } else {
+    res.json(titlesByYear)
+  }
+})
```
idsintehittapa, line 88: This works perfectly! Got the error message 😄

server.js
```diff
@@ -27,9 +19,84 @@ const app = express()
...
+
+// ID-ENDPOINT: to get one title by id (show_id)
+app.get('/netflix-titles/id/:id', async (req, res) => {
```
idsintehittapa, line 91: This was hard to grasp. The database has 2 keys with id in it.
_id  and show_id, I got it wrong the first time trying to find a single item. 

It seems a bit redundant to have it as` /id/:id ` - consider just having `/:id ` instead 😄

server.js
```diff
@@ -27,9 +19,84 @@ const app = express()
...
+  const { id } = req.params;
+  const singleTitleById = await NetflixTitle.findOne({ show_id: id })
+
+  if (singleTitleById) {
+    res.json(singleTitleById)
+  } else {
+    res.status(404).json({ error: `Could not find any Netflix-title with id: ${id}` })
+  }
```
idsintehittapa, line 99: Works! Nice ^^

idsintehittapa (approved):
Hi Hanna!

You’ve made a nice API :) RegExp was new to me so that was cool! 

I’ve left some suggestions. After digging deeper into your API I think documentation is very valuable. I did not include one this time either - but I’ve now learned that it’s a very good practice and nice for the user :D 

Ps, you are sometimes mixing the usage of ; and not (I think you are using ; but in the code that was part of the project there was none, but check it)

server.js
KarolCislowski: when you are filtering anything else then `id` you need to add the name of a filtered field to the path so `/netflix-titles/year/:year` is the good endpoint.

KarolCislowski (approved):
Very nice Project 😸

PR #124, opened 2020-12-20
Agnes top-music

aagneshansson:


server.js
```diff
@@ -27,12 +28,132 @@ const app = express()
...
+  const { name } = req.params;
+  const singleName = await Tracks.findOne({ trackName: name })
+  res.json(singleName);
+});
+
+// Limiting returning data
+app.get('/tracks20', async (req, res) => {
+  const allTracks = await Tracks.find(req.query).skip(20).limit(20);
+  res.json(allTracks);
+});
```
Gabbi-89, line 135: I just wonder why you use .skip(20), I think it is enough with .limit(20) otherwise you skip the 20 first tracks and start from track number 21. Maybe you also can add some sorting here? Like sorting on reviews and show the 20 most popular tracks or anything else? I think that makes it more useful for the user :)

server.js
```diff
@@ -27,12 +28,132 @@ const app = express()
...
+app.get('/artists/:id/tracks', async (req, res) => {
+  const artist = await Tracks.findById(req.params.id);
+
+  if (artist) {
+    const tracks = await Tracks.find({ artistName: mongoose.Types.ObjectId(artist.id) });
+    res.json(tracks);
+  } else {
+    res.status(404).json({error: 'Artist not found!' })
+  }
+});
```
Gabbi-89, line 154: Very impressive that you managed to connect two collections!! 💯 ⭐

server.js
Gabbi-89: But very good that you have tried the mongoose methods :)

Gabbi-89 (changes requested):
Really impressive that you have connected to tracks and artists collections and created some different endpoints! ⭐ The code is clean and you have used the commenting in a good and useful way. You have absolutely reached the requirements and now you can develop even further with your backend coding ;)

You have used try and catch for you first endpoint, maybe try to implement it for all?

Next time: Try  to use query parameters as well for more sorting in the data ;)

I will continue the reviewing when your API link works, right now it says "Service unavailable".. :/

server.js
```diff
@@ -27,12 +28,132 @@ const app = express()
...
+  } else {
+    res.status(404).json({ error: 'Could not find this track' });
+  }
+} catch (error) {
+  res.status(400).json({ error: 'Invalid track id' });
+}
+});
+
+// Route to search by tracknamr
+app.get('/tracks/:name', async (req, res) => {
```
Gabbi-89, line 125: Since the one for :id already are a search field after /tracks/ I cannot search on trackname. I am doing wrong in the search? Otherwise please rename the enpoint or create a query of it instead 😃

Gabbi-89 (approved):

server.js
KarolCislowski, line 125: to make it work you need to make this endpoint URL `/track/name/:name` because `/tract/:name` will start as `track/:id` which is earlier in your `server.js`.

KarolCislowski (approved):
Very nice project 😺

PR #128, opened 2020-12-20
Project-mongo-api - Sara Heijkenskjöld

sarahejhej:


server.js
```diff
@@ -1,38 +1,225 @@
...
+// /eruptions/:id endpoint
+// RETURNS: One unique item
+//
+// PARAMETERS:
+// - id: a unique number
+//    usage: /eruptions/390847
+//
+app.get('/eruptions/:id', async (request, response) => {
+  try {
+    const singleVolcano = await Eruption.findById(request.params.id);
```
Hanna-Skeppe, line 143: Is it a single volcano or a single eruption?

Hanna-Skeppe (approved):
Hej Sara! I did not have very much to comment on other than I think you did a great project. Indentations, spaces and semicolons seems to be in place as far as I can tell🌟.  It was a nice touch that you found your own data to use, which made it more interesting to review. 

It seems like you fulfilled general requirements and also the red level and some of the black. You also used .populate that I was not familiar with so that was interesting. 

I had a few comments you can take a look at but all in all nice job!🌼

KarolCislowski (approved):
Great Project 💯

PR #133, opened 2021-01-10
Project-mongo-api-Claudia 

ckjerrgren:
Had a whole bunch of technical errors. That's why im handing it in super late!

### DIFF
--- a/server.js
+++ b/server.js
@@ -17,49 +17,106 @@ app.use(cors())
 app.use(bodyParser.json())
 
 
-const Globe = new mongoose.model('Globe', {
+const Nomination = new mongoose.model('Nomination', {
     year_film: Number,
     year_award: Number,
     ceremony: Number,
     category: String,
     nominee: String,
     film: String,
-    win: {
-      type: Boolean,
-      required: true 
-    } 
+    win: Boolean, 
 });
 
 
 if (process.env.RESET_DATABASE) {
 
   const populateDatabase = async () => {
-    await Globe.deleteMany();
+    await Nomination.deleteMany();
 
     goldenGlobesData.forEach(item => {
-      const newGlobe = new Globe(item)
-      newGlobe.save();
+      const newNomination = new Nomination(item)
+      newNomination.save();
     })
   }
   populateDatabase();
 }
 
+app.use((req, res, next) => {
+  if (mongoose.connection.readyState === 1) {
+    next()
+  } else {
+    res.status(503).json({ error: 'Service unavailable '})
+  }
+})
+
 // Start defining your routes here
+// 1
 app.get('/', (req, res) => {
   res.send('Hello Internet')
 })
 
-app.get('/globes', async (req, res) => {
-  const allGlobes = await Globe.find(req.query);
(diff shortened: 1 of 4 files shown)


server.js
```diff
@@ -3,36 +3,63 @@ import bodyParser from 'body-parser'
...
+const Globe = new mongoose.model('Globe', {
+    year_film: Number,
+    year_award: Number,
+    ceremony: Number,
+    category: String,
+    nominee: String,
+    film: String,
+    win: {
+      type: Boolean,
+      required: true
```
ellenbjo, line 29: Out of curiosity - why is this required?

server.js
```diff
@@ -3,36 +3,63 @@ import bodyParser from 'body-parser'
...
+    await Globe.deleteMany();
+
+    goldenGlobesData.forEach(item => {
+      const newGlobe = new Globe(item)
+      newGlobe.save();
+    })
+  }
+  populateDatabase();
+}
+
```
ellenbjo, line 46: well done seeding the database!🌟

server.js
```diff
@@ -3,36 +3,63 @@ import bodyParser from 'body-parser'
...
 // Start defining your routes here
 app.get('/', (req, res) => {
-  res.send('Hello world')
+  res.send('Hello Internet')
+})
+
+app.get('/globes', async (req, res) => {
```
ellenbjo, line 52: Maybe the naming could be more specific here, to make it more RESTful. Not sure what globes is here? All nominations? :)

ellenbjo (changes requested):
Hi Claudia,  well done! 🌟 Sorry to hear about all the technical issues you've been having.

I'm going to mark this project with changes requested since I can't find any link to a deployed project? This made it a bit difficult to see what data that was presented in the different endpoints. But other than that i think all of the other blue requirements are met!😊

server.js
```diff
@@ -3,36 +3,63 @@ import bodyParser from 'body-parser'
...
+  const allGlobes = await Globe.find(req.query);
+  res.json(allGlobes);
+})
+
+app.get('/globes/:nominee', async (req, res) => {
```
KarolCislowski, line 57: `nominee` is not a good key for extracting single document, unfortunately. When you use it like this you returning a bad set of data because you cutting off a big portion of it. For example "Avatar" has a lot of nominations, but you are showing only the first one.

KarolCislowski (changes requested):
Please add the good endpoint to a single document.

server.js
ckjerrgren: Changed it to a find so it shows all the nominations an actor or movie have.

KarolCislowski: If you change this one to `find()` it will make your data complete, but you still need to provide the endpoint to the single element by using `_id` as a key

ckjerrgren: That specific data  didn't have an Id key. 
But I tried solving it using multiple paths..

ckjerrgren:
Added more endpoints and paths. 

I know I sometimes use Nomination to get data and sometimes use goldenGlobesData... I'm not 100% on why that happened.

PR #134, opened 2021-01-11
Axel Hammarbäck – Project Mongo API

AxelHammarback:
Sorry for late pull request, was sure I'd done it already


server.js
```diff
@@ -3,35 +3,97 @@ import bodyParser from 'body-parser'
...
+  // These are the properties we're filtering on. All three in one go, yeeeah!
+  let filteredNominees = await Nominee.find(
+  { 
+    year_award: year, 
+    category: category, 
+    win: true 
+  })
+  res.json(filteredNominees)
+})
+
```
lilihorvath, line 96: I can not find any of mistake, it seems just perfect for me. We talked about the read.me file, but that is just a little thing. Your could is much more understandable and readable than in the lecture was so big congrats Axel. I thought that I'll write you many of comment just as I used to for others, but this is a very good job! And also I have learnt so many new examples from your code! Thank you!

server.js
```diff
@@ -3,35 +3,97 @@ import bodyParser from 'body-parser'
...
   res.send('Hello world')
 })
 
+// ⤵ Endpoint that returns all the nominees, ever. Also works as a query param.
+app.get('/nominees', async (req, res) => {
+  // Whoa! The little "req.query" in the find() function works as a query param. So, you can for example type /nominees?nominee=Kaitlyn%20Dever and learn that Kaitlyn Dever was nominated for "Best Performance by an Actress in a Limited Series or a Motion Picture Made for Television" in 2020, but sadly didn't win (Michelle Williams won).
+  const allNominees = await Nominee.find(req.query)
+  res.json(allNominees)
+})
+
```
KarolCislowski, line 59: It would be nice to have `/nominees/:id` to retrieve a single document easy way

data/technigo-staff.json
```diff
@@ -0,0 +1,30 @@
+[
```
KarolCislowski, line 1: And this file with personal data is created here for?

KarolCislowski (approved):
Great work 💯

PR #156, opened 2021-05-15
Project Mongo API - Åsa Hildebrand

AsaHildebrand:
I think this project went well, aiming for the blue goals.


server.js
```diff
@@ -1,35 +1,91 @@
...
 
 // Start defining your routes here
 app.get('/', (req, res) => {
-  res.send('Hello world')
+  res.send(listEndpoints(app))
+})
+
+// Endpoint that shows all books
+app.get('/books', async (req, res) => {
+  const books = await Book.find(req.query)
```
katliebig, line 58: req.query is not needed here, we just assume that it's left from trying to add a query

server.js
```diff
@@ -1,35 +1,91 @@
...
+    const books = await Books.find({
+      author: {
+        $regex: new RegExp(author, "i")
+      }
+    })
+    res.json(books)
+  } else {
+    const books = await Books.find()
+    res.json(books)
+  }
```
katliebig, line 76: You could move res.json(books) to after the if, because you want to send that in both cases. One line less! 😎

server.js
```diff
@@ -1,35 +1,91 @@
...
+
+// Endpoint that shows books by id (ex: books/9)
+app.get('/books/:id', async (req, res) => {
+  const { id } = req.params
+  const singleBook = await Book.findOne({ bookID: +id })
+
+  if (singleBook) {
+    res.json(singleBook)
+  } else {
+    res.status(404).json({ error: 'Could not find this book'})
```
katliebig, line 87: Nice error handling!

server.js
```diff
@@ -1,35 +1,91 @@
...
+  } else {
+    const books = await Books.find()
+    res.json(books)
+  }
+})
+
+// Endpoint that shows books by id (ex: books/9)
+app.get('/books/:id', async (req, res) => {
+  const { id } = req.params
+  const singleBook = await Book.findOne({ bookID: +id })
```
katliebig, line 82: Here it might be useful to specify which ID is being used (bookID), because Mongo also makes its own id, (_id). The latter one would work with Book.findById(id).

katliebig (approved):
Good work! Clean and readable code 😎
One suggestion is to add try catch to all async functions to catch unforeseen errors. For example, we broke your application by trying to search for a book with the ID created by Mongo (https://week18project-mongo-api.herokuapp.com/books/609eb9ead721660028f3e722).
With try catch, you can catch this and provide the frontend with a 400 (bad request) status code and a general message.

README.md
```diff
@@ -1,13 +1,11 @@
...
 
 ## The problem
 
-Describe how you approached to problem, and what tools and techniques you used to solve it. How did you plan? What technologies did you use? If you had more time, what would be next?
+This week I didn't have so many error messages as I had last week so it felt a bit easier to work with this API project. I watched the lectures and tried to implement different endpoints and play around with it to learn more. It was a bit difficult to deploy it to Heruko and Mongo but it turned out well after help from my team and discussions on Slack.
 
 ## View it live
 
-Every project should be deployed somewhere. Be sure to include the link to the deployed project so that the viewer can click around and see what it's all about.
+https://week18project-mongo-api.herokuapp.com/
```
KarolCislowski, line 11: short documentation would be nice 🤓

server.js
```diff
@@ -1,39 +1,95 @@
...
+const mongoUrl = process.env.MONGO_URL || "mongodb://localhost/mongo-books"
 mongoose.connect(mongoUrl, { useNewUrlParser: true, useUnifiedTopology: true })
 mongoose.Promise = Promise
 
-// Defines the port the app will run on. Defaults to 8080, but can be 
-// overridden when starting the server. For example:
-//
-//   PORT=9000 npm start
+const bookSchema = new mongoose.Schema({
+  bookID: Number,
```
KarolCislowski, line 17: you will have autogenerated _id field so it is no need to pass the old id 🤓

server.js
```diff
@@ -1,39 +1,95 @@
...
+    const books = await Books.find({
+      author: {
+        $regex: new RegExp(author, "i")
+      }
+    })
+    res.json(books)
+  } else {
+    const books = await Books.find()
+    res.json(books)
+  }
```
KarolCislowski, line 76: nice filtering

server.js
KarolCislowski: it is better to use _id field generated by mongo, because it is indexed, that why it would be much faster on bigger datasets.

KarolCislowski (approved):
Nice API 🥇

PR #172, opened 2021-05-16
Maria Pogosjan - project-mongo-api

MariaPogosjan:
Hi Karol! 

Hope you are doing well and that you will enjoy my code this week ☺️  

I have a problem with my code and it is the "$limt" in line 81 in the code. 
I had a 1:1 with Maks about it because I could not make it work and got the error message _"Limit must be a positive integer"_. But during our 1:1 it worked (strange...). Today when I am trying the live version it stopped working again and had to comment it out. The only thing that did not work is /movies. Could you please look at it and maybe see what the problem can be? 

Bye bye!

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,4 +9,4 @@ This week we started using a database to store and retrieve data from and use th
 
 ## View it live
 
-Live 👉🏻  https://poggi-project-mongo.herokuapp.com/
+Live 👉🏻   https://poggi-project-mongo.herokuapp.com/
--- a/README.md
+++ b/README.md
@@ -1,13 +1,12 @@
 # Mongo API Project
+This week we started using a database to store and retrieve data from and use that data to produce a RESTful API. We also modeled our own database using Mongoose models, and persist our data in the database. I have used 1375 Netflix titles dataset. 
 
-Replace this readme with your own information about your project. 
+## What I have learned 🧠
 
-Start by briefly describing the assignment in a sentence or two. Keep it short and to the point.
-
-## The problem
-
-Describe how you approached to problem, and what tools and techniques you used to solve it. How did you plan? What technologies did you use? If you had more time, what would be next?
+- How to model data in Mongoose
+- How to fetch items from a Mongo database using Mongoose
+- How to seed large amounts of data to a database
 
 ## View it live
 
-Every project should be deployed somewhere. Be sure to include the link to the deployed project so that the viewer can click around and see what it's all about.
+Live 👉🏻  https://poggi-project-mongo.herokuapp.com/
--- a/README.md
+++ b/README.md
@@ -7,6 +7,15 @@ This week we started using a database to store and retrieve data from and use th
 - How to fetch items from a Mongo database using Mongoose
 - How to seed large amounts of data to a database
 
+## Endpoints
+- GET '/': returns the list of all endpoints;
+- GET '/movies': returns a list of all movie titles and query params for country, cast, show type, pagination for pages;
+- GET '/movies/titles': returns only titles sorted alphabetically;
+- GET '/movies/genres': returns all diffrent genres; 
+- GET '/movies/years': returns all the diffrent years; 
+- GET '/movies/years/latest': returns only movies from year 2018 until now;
+- GET '/movies/:movieId': returns a particular movie;  
+
 ## View it live
 
 Live 👉🏻   https://poggi-project-mongo.herokuapp.com/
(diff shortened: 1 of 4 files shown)


server.js
```diff
@@ -1,38 +1,161 @@
...
+      $match: {
+        country: {
+          $regex: new RegExp(country, "i")
+        }, 
+        cast: {
+          $regex: new RegExp(cast, "i")
+        },
+        type: {
+          $regex: new RegExp(show, "i")
+        }
```
Martno96, line 74: I'm really impressed with your use of mongoose queries :D

server.js
```diff
@@ -1,38 +1,161 @@
...
-app.use(bodyParser.json())
+app.use(express.json())
 
-// Start defining your routes here
 app.get('/', (req, res) => {
-  res.send('Hello world')
+  res.send(listEndpoints(app))
+})
+
+app.get('/movies', async (req, res) => {
```
Martno96, line 54: To me, this endpoint works just fine :D

server.js
```diff
@@ -1,38 +1,161 @@
...
+    if (!yearUnique.includes(item)) {
+      yearUnique.push(item)
+    }
+    return yearUnique.sort().reverse()
+  })
+
+  res.json({ length: yearUnique.length, data: yearUnique })
+})
+
+app.get('/movies/years/latest', async (req, res) => {
```
Martno96, line 130: The data structure for this endpoint differs a bit from what I expected :O Since (I’m assuming) the parent endpoint in the hierarchy (/movies/years) is intended to include movies per year, I expected this endpoint to do the same, just that it fetches only recent years. Also, the movies aren’t currently sorted by release year, which too would be solved by "inheriting" the data structure of /movies/years :)

Martno96 (commented):
Good job Maria! :D I'm impressed with how ambitious you've been with the use of mongoose queries. Unfortunately I'm not yet of much help in that respect, having not had time to really get into the area yet. However, I do think I have a clue as to why the specific movie endpoint isn't working, and I left some suggestions on alternative data structure for two endpoints :D All in all, keep up the good work!

server.js
```diff
@@ -1,38 +1,161 @@
...
+        type: {
+          $regex: new RegExp(show, "i")
+        }
+      }
+    }, 
+    {
+      $skip: Number((page + 1) * per_page + 1)
+    }/* , 
+    {
+      $limit: Number(per_page)
```
KarolCislowski, line 81: you need to provide the default `per_page` value because here you trying parsing undefined into the number if your user will not provide you with value in the query. 

```terminal
> Number(undefined)
< NaN
```

server.js
```diff
@@ -1,38 +1,161 @@
...
+app.use(express.json())
 
-// Start defining your routes here
 app.get('/', (req, res) => {
-  res.send('Hello world')
+  res.send(listEndpoints(app))
+})
+
+app.get('/movies', async (req, res) => {
+  const { country, cast, show, page, per_page } = req.query
```
KarolCislowski, line 55: `const { country, cast, show, page, per_page=20 } = req.query`

server.js
KarolCislowski, line 74: nice filtering 👍🏻

README.md
```diff
@@ -1,13 +1,12 @@
...
-
-Describe how you approached to problem, and what tools and techniques you used to solve it. How did you plan? What technologies did you use? If you had more time, what would be next?
+- How to model data in Mongoose
+- How to fetch items from a Mongo database using Mongoose
+- How to seed large amounts of data to a database
 
 ## View it live
 
-Every project should be deployed somewhere. Be sure to include the link to the deployed project so that the viewer can click around and see what it's all about.
+Live 👉🏻   https://poggi-project-mongo.herokuapp.com/
```
KarolCislowski, line 12: short documentation would be nice 😸

server.js
```diff
@@ -1,38 +1,161 @@
...
+  ]) 
+
+  res.json({ length: movies.length, data: movies })
+})
+
+app.get('/movies/titles', async (req, res) => {
+  const title = await Movie.aggregate([
+    { $sort: { title: 1 } },
+    { $unset: ["_id", "director", "cast", "country", "date_added", "release_year", "rating", "duration", "listed_in", "description", "type", "__v"] }
+  ])
```
KarolCislowski, line 92: great usage of the aggregate() 🥳

server.js
KarolCislowski: When something is yealing about value, remember to console.log value the input before and after the function. 

```js
console.log(typeof(per_page))
const numberPefPage = Number(per_page)
console.log(typeof(numberPerPage))
```

and then you would see 
```
< undefined
< NaN
```

KarolCislowski (approved):
Great API 👍🏻

PR #179, opened 2021-05-16
Project mongo-api Lars Söderman

lars-soderman:
Here it is! Very basic project this week, but hope you like!


server.js
```diff
@@ -25,12 +52,22 @@ const app = express()
...
+app.get('/books', (req, res) => {
+  Book.find().then((data) => res.json(data))
+})
+app.get('/books/:ID', async (req, res) => {
+  const { ID } = req.params
+  if (ID) {
+    const response = await Book.find({ bookID: ID })
+    res.json(response)
+  }
+})
```
magicvilla, line 71: Nice endpoints and really good job with your path param to getting the original book id! 💯 
It would have been nice to have som error responses e.g if the user enters an id that does not match or if the server is down for some reason. Otherwise really nice endpoints and working like a charm 🧙

server.js
```diff
@@ -1,21 +1,48 @@
...
+  isbn13: Number,
+  average_rating: Number,
+  language_code: String,
+  num_pages: Number,
+  ratings_count: Number
+
+});
+
+const Book = mongoose.model('Book', bookSchema);
+
```
magicvilla, line 24: Nice structuring of your book schema! 🥇

magicvilla (approved):
Good job Turtle Lars 😄 You did good work on this weeks project! Your api works well and you have some nice endpoints! I good tip is to add som error responses to if the server would be down or if for some reason wrong id for books is entered! Also remember to be consistent with semi-colons in the code to clean up a bit. I always tend to forget too! 💯

server.js
```diff
@@ -1,21 +1,48 @@
...
-// import netflixData from './data/netflix-titles.json'
-// import topMusicData from './data/top-music.json'
+import booksData from './data/books.json'
 
 const mongoUrl = process.env.MONGO_URL || "mongodb://localhost/project-mongo"
 mongoose.connect(mongoUrl, { useNewUrlParser: true, useUnifiedTopology: true })
 mongoose.Promise = Promise
 
+const bookSchema = new mongoose.Schema({
+  bookID: Number,
```
KarolCislowski, line 12: why you need both bookId and _id??

KarolCislowski (approved):
Nice API 👍🏻

PR #182, opened 2021-05-16
Project-mongo-api / Honkonen Anna-Kaisa

honkoan:
This week I did only the basics. I would have liked to do also the frontend but I didn't have enough time. 

https://exercises-api-mongodb.herokuapp.com/


.gitignore
```diff
@@ -1,5 +1,6 @@
 node_modules
 .DS_Store
+.env
```
ThereseHag, line 3: Great that you remembered to put this in the gitignore file.

README.md
```diff
@@ -1,13 +1,48 @@
...
+/exercise?name=deadlift  
+
+* how to find all exercises with target muscle gluteus maximus:  
+/exercise?targetMuscle=gluteus 
+
+### All multi-joint exercises  
+/category/multi-joint
+
+### All single-joint exercises  
+/category/single-joint
```
ThereseHag, line 26: I really like that you put your API documentation here, where it's easy to find. Also, great job that you did a documentation 🌟

server.js
```diff
@@ -1,35 +1,107 @@
...
 
-// If you're using one of our datasets, uncomment the appropriate import below
-// to get started!
-// 
-// import goldenGlobesData from './data/golden-globes.json'
-// import avocadoSalesData from './data/avocado-sales.json'
-// import booksData from './data/books.json'
-// import netflixData from './data/netflix-titles.json'
-// import topMusicData from './data/top-music.json'
+import exercisesData from './data/exercises.json'
```
ThereseHag, line 8: I really like the fact that you're taking the time to create a dataset of your own. Great job!

server.js
```diff
@@ -1,35 +1,107 @@
...
-// Add middlewares to enable cors and json body parsing
 app.use(cors())
-app.use(bodyParser.json())
+app.use(express.json())
 
-// Start defining your routes here
+// Endpoints
 app.get('/', (req, res) => {
-  res.send('Hello world')
+  res.send(listEndpoints(app))
```
ThereseHag, line 48: It's nice that you use this to display the endpoints on the first page. It gives a nice overview.

server.js
```diff
@@ -1,35 +1,107 @@
...
+// All exercises
+app.get('/exercises', (req, res) => {
+  Exercise.find().then((exercises) => {
+    res.json({ data: exercises })
+  })
+})
+
+// Find exercise by id
+app.get('/exercises/:exerciseId', async (req, res) => {
+  const { exerciseId } = req.params
```
ThereseHag, line 60: Here I think it would be better to use the automatic id that is created in the database (the one that is called _id) since that is always guaranteed to be unique.

server.js
```diff
@@ -1,35 +1,107 @@
...
+
+  const exercise = await Exercise.aggregate([
+    {
+      $match: {
+        name: {
+          $regex: new RegExp(name || "", "i")
+        },
+        target_muscle: {
+          $regex: new RegExp(targetMuscle || "", "i")
+        }
```
ThereseHag, line 82: Really nice that you used this. I find it very useful and can't wait to use it in larger projects :)

ThereseHag (approved):
You've done a great job here! I especially appreciate that you have created your own dataset that you're working with, and that you have included a documentation of that API in your Readme, which makes it very easy to get an idea of how it can be used. 

If I were to say anything as an advice on the way; then it would be to remember to use the unique id created in the database, instead of the one you've created yourself in the dataset. That way, it will be safer and you can always be sure that every id is really unique.

README.md
KarolCislowski, line 26: Great documentation 👍🏻

server.js
```diff
@@ -1,35 +1,107 @@
...
+const mongoUrl = process.env.MONGO_URL || 'mongodb://localhost/project-mongo'
 mongoose.connect(mongoUrl, { useNewUrlParser: true, useUnifiedTopology: true })
 mongoose.Promise = Promise
 
-// Defines the port the app will run on. Defaults to 8080, but can be 
-// overridden when starting the server. For example:
-//
-//   PORT=9000 npm start
+const exerciseSchema = new mongoose.Schema({
+  exerciseID: Number,
```
KarolCislowski, line 17: Why do you need both exerciseID and _id?

server.js
KarolCislowski, line 82: Nice filtering 👍🏻

KarolCislowski (approved):
Great API 👍🏻

PR #191, opened 2021-05-20
Project Mongo-Api Maidelin rubio

rubmaigut:


server.js
```diff
@@ -1,35 +1,88 @@
...
 const mongoUrl = process.env.MONGO_URL || "mongodb://localhost/project-mongo"
 mongoose.connect(mongoUrl, { useNewUrlParser: true, useUnifiedTopology: true })
 mongoose.Promise = Promise
 
-// Defines the port the app will run on. Defaults to 8080, but can be 
-// overridden when starting the server. For example:
-//
-//   PORT=9000 npm start
+const Serie= mongoose.model('Series', {
+  show_id: Number,
```
KarolCislowski, line 12: Why do you need both show_id and _id?

README.md
```diff
@@ -1,13 +1,19 @@
...
-Describe how you approached to problem, and what tools and techniques you used to solve it. How did you plan? What technologies did you use? If you had more time, what would be next?
+I couldn't advance too much in this project. I only got blue requirements following the same structure that we used the last week.
+
+I hope that for the next week I can learn more about Mongo and the best way to do it. 
+
+The API has the following endpoints:
+
+/showId ---- To filter a specific ID
+/title/ShowTitle ----- to search by titles
+/type/type ---- to filter by type (movies - Tv Show)
```
KarolCislowski, line 15: /series/... You need to correct your endpoints documentation.

KarolCislowski (approved):
Nice API 👍🏻

server.js
rubmaigut: HI, @KarolCislowski I decided to use show_id that is the original id of the series as this data was converted from Json to mongo I didn't use the ID generated when it creates a new document in de DB because was too long and to understand better in this project I kept the original id.

README.md
rubmaigut: It's updated thank you!

PR #225, opened 2021-12-18
Patrik - Project mongo api

Mattsson717:


server.js
```diff
@@ -1,38 +1,130 @@
...
+});
+
+// Get all the books
+app.get("/books", async (req, res) => {
+  let books = await Book.find(req.query);
+
+  if (req.query.numPages) {
+    const booksByPages = await Book.find().gt("numPages", req.query.numPages);
+    books = booksByPages;
+  }
```
katiewu1, line 70: Maybe can throw in an error handling here? For example when I have url https://mattssons-book-api.herokuapp.com/books?numPages=10000, I get an empty array.

server.js
```diff
@@ -1,38 +1,130 @@
...
+  res.json(books);
+});
+
+app.get("/books/:id", async (req, res) => {
+  try {
+    const bookById = await Book.findById(req.params.id);
+    if (bookById) {
+      res.json({
+        response: bookById,
+        succes: true,
```
katiewu1, line 80: typo

server.js
```diff
@@ -1,38 +1,130 @@
...
+      });
+    } else {
+      res.status(404).json({
+        response: "Book not found",
+        succes: false,
```
katiewu1, line 85: typo

server.js
```diff
@@ -1,38 +1,130 @@
...
+      response: "Invalid id",
+      success: false,
+    });
+  }
+});
+
+// Get authors
+// app.get("/books/authors/:author", async (req, res) => {
+//   const bookByAuthor = await Book.find({ authors: req.params.authors });
+//   if (bookByAuthor) {
```
katiewu1, line 99: I think this endpoint with the :author placeholder is better than having only Tolkien endpoint from line 113. If we have a placeholder then it's up to the frontend to decide which author they want to retrieve data.

server.js
```diff
@@ -1,38 +1,130 @@
...
+
+// Get all the books
+app.get("/books", async (req, res) => {
+  let books = await Book.find(req.query);
+
+  if (req.query.numPages) {
+    const booksByPages = await Book.find().gt("numPages", req.query.numPages);
+    books = booksByPages;
+  }
+  res.json(books);
```
katiewu1, line 71: The response here is different if you compare to the response from line 78-81. I think it's better to have a consistent way so it's easier for the frontend to understand this API. Otherwise they have to code differently when they want to retrieve data from each response. For example on line 79 you need to fetch the data with a .response but not here on this response.

server.js
```diff
@@ -1,38 +1,130 @@
...
+// Get all books by J.R.R Tolkien
+app.get("/books/authors/Tolkien", async (req, res) => {
+  await Book.find({ authors: "J.R.R. Tolkien" }, (error, data) => {
+    if (error) {
+      res.status(404).json({
+        response: "Author not found",
+        succes: false,
+      });
+    } else {
+      res.send(data);
```
katiewu1, line 121: The response here differ too

katiewu1 (commented):
Hej Patrik, great job with this backend week! 
I made a few comments, hope you find them useful.
And I notice that you have a data/games.json file that you don't use? Maybe can delete it? It's quite a large file :)

PR #237, opened 2021-12-23
Hedvigs mongo api

HedvigM:
# Mongo API Project

This is my second API project. I used music data to make some endpoints using mongoose.

## Features

- 4 different endpoints using mongoose methods like find, findById, and gt.
- Endpoints that return a collection of results: endpoints, tracks, and danceability.
- error handling by returning 404 and 400 status and an error message.

## The problem

If I had more time I would get another data set to work with and make a front to my backend. I would love to pagination, but that's for the next project.

## View it live

https://hedvigs-second-api.herokuapp.com

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,12 +1,16 @@
 # Mongo API Project
 
-Replace this readme with your own information about your project. 
+This is my second api project. I used music data to make some endpoints using mongoose.
 
-Start by briefly describing the assignment in a sentence or two. Keep it short and to the point.
+## Features
+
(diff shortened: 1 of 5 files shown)


IdaAspen (approved):
Well done Hedvig! You have fulfilled all the blue requirements.
The code is clean and easy to read. Maybe instead of describing the endpoints in server.js you can add a description to them in a separate document or in the Readme?
 I haven't much else to add. It looks great! 🎉

PR #240, opened 2022-01-02
Week 18 project pull request from Maria Sjögren

molbimien:
Here's my pull request for this week's project:  Build a Mongo-backed API

## What?

This week I kept it really simple and only did what was needed to reach the blue requirements. When I have more time, I will restart this project and grab some dataset from Kaggle and then build a frontend to go with the backend. 

## Find the site here:

https://molbimien-project-mongo-api.herokuapp.com

### DIFF
--- a/server.js
+++ b/server.js
@@ -2,6 +2,8 @@ import express from 'express'
 import cors from 'cors'
 import mongoose from 'mongoose'
 
+import netflixData from './data/netflix-titles.json'
+
 // If you're using one of our datasets, uncomment the appropriate import below
 // to get started!
 // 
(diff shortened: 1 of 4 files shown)


server.js
```diff
@@ -26,13 +21,81 @@ const app = express()
...
+    })
+  }
+  
+  seedDatabase()
+}
+
 // Start defining your routes here
 app.get('/', (req, res) => {
-  res.send('Hello world')
+  // res.send('Hello from us!')
```
Skrosen, line 65: Remember to erase this if you don't need it :)

server.js
```diff
@@ -26,13 +21,81 @@ const app = express()
...
+    } else {
+      res.json({
+        response: netflixTitleById,
+        success: true,
+      })
+    }
+  } catch (err) {
+    res.status(400).json({
+      error: 'Id is invalid'
+    })
```
Skrosen, line 93: I think you can use the fact that you wrote (err) and send it in row 92 as "error: err" and you will get more information from the error message (I haven't tried it out on your code but just realised yesterday how this actually works so could be worth looking at what you get if you send the whole err) :) You can send "message: 'Id is invalid" if you want to as well!

Skrosen (commented):
Good job! The way you've constructed your API is restful and you reach the blue requirements.

PR #258, opened 2022-05-11
Lisas project-mongo-api

lisabergstrom:


EmmaaBerg (commented):
Good job, the project fulfill the blue requirements. 
If you would like maybe you could add some statuses to show the user when it was a success or not a successfull response.

PR #262, opened 2022-05-12
project-mongo-api-frida

Nosslexa:
View it live: https://project-mongo-api-frida.herokuapp.com/

### DIFF
--- a/server.js
+++ b/server.js
@@ -28,12 +28,12 @@ const Book = mongoose.model("Book", {
 if (process.env.RESET_DB) {
   const seedDatabase = async () => {
     await Book.deleteMany();
-    booksData.forEach( singleBook => {
+    booksData.forEach((singleBook) => {
       const newBook = new Book(singleBook);
       newBook.save();
-    })
+    });
   };
-  seedDatabase()
+  seedDatabase();
 }
 
 app.use(cors());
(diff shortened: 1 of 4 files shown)


server.js
```diff
@@ -1,32 +1,65 @@
...
+// 2. endpoint that returns one rated book by its title.
+app.get("/bookratings/booktitle/:title", async (req, res) => {
+  const bookByTitle = await Book.findOne({ title: req.params.title });
+  res.send(bookByTitle);
+});
+
+// // 3. endpoint that returns all rated books written by a specific author.
+app.get("/bookratings/author/:authors", async (req, res) => {
+  const bookByAuthor = await Book.find({ authors: req.params.authors });
+  res.send(bookByAuthor);
```
rawisou, line 62: Great job returning an array of (multiple) books by one author.
The only suggestion I have would be to allocate for case sensitivity.
In this case, the easiest would be to use` RegExp`

For example: 
Line 61, we could do something like this:

`const bookByAuthor = await Book.find({ 
authors: new RegExp (req.params.authors, 'i')
});`

Or save req.params.authors in a variable and replace it with the variable like:

```
const findByAuthors = req.params.authors
const bookByAuthor = await Book.find({ 
authors: new RegExp (findByAuthors, 'i')
});
```
We can also destructure the req.params instead of a new variable

```
const { authors } = req.params
const bookByAuthor = await Book.find({ 
authors: new RegExp (authors, 'i')
});
```

rawisou (commented):
Hi Frida,

Your code is easy to follow and you have listed all the endpoints in the base URL which is a nice touch. The endpoints work as intended.

I have left a comment as an optional addition to your already-well-written code. :)

PR #266, opened 2022-05-13
Project Mongo API Hanna

HWallberg:
https://project-mongo-api-hanna.herokuapp.com/


Katarina821 (approved):
Hi Hanna! Nice job on the project, you endponts are working, and also a nice solution with the regex! Your code is well strcutured, nicely done!

PR #274, opened 2022-05-13
Project Mongo API – Emma Lindell week 18

emmajosefina:
https://emmas-songs.herokuapp.com/songs


server.js
```diff
@@ -1,35 +1,95 @@
...
+  try {
+    let { page, size } = req.query
+    if (!page) {
+      page = 1
+    } if (!size) {
+      size = 100
+    }
+
+    const limit = parseInt(size, 100)
+    const skip = (page - 1) * size
```
Neaa99, line 72: What does this do? I havn't seen it before, looks intresting

server.js
```diff
@@ -1,35 +1,95 @@
...
+  } catch (error) {
+    res.status(400).json({ error: "Error, not a page or size"})
+  }
+})
+
+
+// Artist name
+app.get("/songs/song/:artistName", async (req, res) => {
+  // this will retrieve only the first found song
+  const singleSong = await Song.findOne({artistName: req.params.artistName})
```
Neaa99, line 85: Insted of findOne you can just use find since the are artist with multible songs. Shawn Mendes for example has two songs on the list but only the firt one comes up. So maybe take a look at that

server.js
```diff
@@ -1,35 +1,95 @@
...
+  res.send(singleSong)
+})
+
+
 
 // Start the server
 app.listen(port, () => {
-  console.log(`Server running on http://localhost:${port}`);
-});
+  console.log(`Hello world`)
```
Neaa99, line 93: Maybe delete this

Neaa99 (commented):
Great project Emma! Your code looks clean and structured and it's easy to read and understand. The only thing I'm wondering is how I can use your media query. I can see that you have a query in line 64, but I don't manage to figure out how to apply it. Maybe clear that up on the start page :)

Good work!

server.js
JenFi72, line 85: I'd like to have some error codes related to the artist search as well. When I type an artist who's name does not exist in the database, I get a blank page. Would've been nice with an error code and message.

JenFi72 (commented):
Great job Emma!!! The endpoints works well and you have used a RESTful approach when setting up the design. I like that you added size to the error code :)

PR #277, opened 2022-05-14
Sofia Ringstedt Mongo API

sofiaringstedt:


ariallahyar (commented):
Nice job this week, Sofia. Your API works as it should. Consider using query parameters when you're designing APIs in the future. I found [this great Stack Overflow answer](https://stackoverflow.com/questions/30967822/when-do-i-use-path-params-vs-query-params-in-a-restful-api) that has helped me approach the setup.

Since you are, in an essence, filtering songs by different parameters, it could make sense to use queries instead. For example, instead of creating specific paths for genre and artist, your path `/songs` would be able to be built upon like this: `/songs?genre=pop&artist=Ed Sheeran`

PR #287, opened 2022-05-15
Project Mango DB

rijad90:


jessand77 (commented):
Good job! I don't have so many comments to leave here but I think the code is clear and easy to follow and that the project fulfills all the general and blue level requirements.

PR #290, opened 2022-05-15
WK18 Mongo API

MarianneArdin:


package.json
```diff
@@ -9,12 +9,21 @@
...
+  },
+  "main": "server.js",
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/MarianneArdin/project-mongo-api.git"
+  },
+  "bugs": {
+    "url": "https://github.com/MarianneArdin/project-mongo-api/issues"
+  },
+  "homepage": "https://github.com/MarianneArdin/project-mongo-api#readme"
```
emmahogberg88, line 28: Cool, I haven't seen this before! what does it do? :)

server.js
```diff
@@ -1,35 +1,110 @@
...
+}
+
 // Add middlewares to enable cors and json body parsing
 app.use(cors());
 app.use(express.json());
 
+// Middleware that checks if the database is connected before going to endpoints
+app.use((req, res, next) => {
+	if (mongoose.connection.readyState === 1) {
+		next();
```
emmahogberg88, line 48: Cool, haven't seen this before either. Have to google 👓

server.js
```diff
@@ -1,35 +1,110 @@
...
+//RESTful routes/endpoints
+app.get('/', (req, res) => {
+	res.send(
+		'Welcome to the Golden-Globes API. Endpoints: /goldenglobes /nominations'
+	);
+});
+
+//this route will display all nominations
+app.get('/nominations', (req, res) => {
+	res.json(data);
```
emmahogberg88, line 63: If you want you can add some error handling here, like you have below. Maybe also add an else if for both success and for error (staus(200) and status(400) for example.

server.js
```diff
@@ -1,35 +1,110 @@
...
+//Shows specific film and if it won or not//
+app.get('/goldenGlobes/win/:showName', (req, res) => {
+	const showName = req.params.showName;
+
+	let data = data.find((item) => item.nominee === showName);
+	res.status(200).json({ data: data.win, success: true });
+});
+
+app.get('/film', async (res, req) => {
+	const film = await film.find({});
```
emmahogberg88, line 95: I cannot test the API so I'm not 100% sure but it seems like you are sending an empty object here?

server.js
```diff
@@ -1,35 +1,110 @@
...
+	res.json(film);
+});
+
+app.get('/goldenGlobes/film/:id/film', async (res, req) => {
+	const film = await film.findById(req.params.id);
+	const nominee = await nominee.find({
+		film: mongoose.Types.ObjectId(film.id),
+	});
+	res.json(nominee);
```
emmahogberg88, line 104: Nice use of async await! Maybe it could be a good idea to use the same syntax for GETting the other endpoints as well :)

README.md
```diff
@@ -1,13 +1,14 @@
...
 
-## The problem
-
-Describe how you approached to problem, and what tools and techniques you used to solve it. How did you plan? What technologies did you use? If you had more time, what would be next?
+I used the basecode from my previous project https://github.com/MarianneArdin/project-express-api
 
 ## View it live
+https://wk18mongoapi.herokuapp.com/
+Endpoints: /goldenglobes
+/nominations
```
emmahogberg88, line 12: Looks like you have more endpoints in your code. Perhaps you can add them here also :)

emmahogberg88 (commented):
Hi Marianne. Great job with your API. Unfortunately, I can't access the API in the browser so I couldn't test your endpoints. I had some issues with this myself. When I got the API to work I got empty arrays back, so I had to add API KEYS in Heroku. This SO question helped me: 
https://stackoverflow.com/c/technigo/questions/2531

Also, I commented out the deleteMany() so the responses wasn't deleted every time i requested data (as you can see in my code if you want to take a look).

However, I have looked through your code and it looks good! You have several endpoints which is nice. You are using some mixing syntax in some places so it might be a good idea to refactor your code so it all follows the syntax as you have for "async await request" in the end, which I have understood is a good syntax to use. Also, you have error handling which is great! I suggest that you use the if else for handling errors for all endpoints (like you did on row 75 to 84). 

Great job!

PR #292, opened 2022-05-15
MongoDb API Emma Springs

EmmaSprings:
Having issues with the heroku site. Will try to solve but code should work still.


Nosslexa (commented):
Great job with this weeks project! 
Your code is easy to read and follow and your routs works as they should. Nice that you have found your own data set to use.

PR #295, opened 2022-05-15
justine-project-mongo-api

JustineZwiazek:


server.js
```diff
@@ -17,16 +13,92 @@ mongoose.Promise = Promise;
...
+    res.json(allLaureates);
+  } catch (error) {
+    res
+      .status(400)
+      .json({ error: "We didn't find what you were looking for." });
+  }
+});
+
+// Req: A minimum of one endpoint to return a single result (single element).
+app.get("/laureates/:id", async (req, res) => {
```
emmajosefina, line 78: This endpoint only works for id: 1 unfortunately, when searching for id 9 or 24 for example, nr 1 still shows up!

emmajosefina (approved):
Good work with last week's project Justine! I had a problem with the id-endpoint, but other than that - everything works great! 💯

PR #298, opened 2022-05-15
Project Mongo API 

idanaslund:
https://project-mongo-api-netflix.herokuapp.com/

### DIFF
--- a/server.js
+++ b/server.js
@@ -72,6 +72,13 @@ app.get("/streams/title/:title", async (req, res) => {
 
 })
 
+app.get("/streams/year/:release_year", async (req, res) => {
+  
+  const releaseYear = await Stream.find({release_year: req.params.release_year})
+  res.send(releaseYear)
+
+})
+
 // Start the server
 app.listen(port, () => {
   console.log(`Server running on http://localhost:${port}`);
--- a/server.js
+++ b/server.js
@@ -1,6 +1,7 @@
 import express from "express";
 import cors from "cors";
 import mongoose from "mongoose";
+import bodyParser from "body-parser"
 import listEndpoints from "express-list-endpoints"
 import netflixData from "./data/netflix-titles.json";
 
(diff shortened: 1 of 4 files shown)


server.js
```diff
@@ -20,15 +15,70 @@ mongoose.Promise = Promise;
...
 
+// When mongoDB is not active
+app.use((req, res, next) => {
+  if (mongoose.connection.readyState === 1) {
+    next()
+  } else {
+    res.status(503).json({ error: "Service unavailable" })
+  }
+})
+
```
CamillaHallberg, line 57: I like that you added the next function in order to show if the service is available or not 👍🏽

server.js
```diff
@@ -20,15 +15,70 @@ mongoose.Promise = Promise;
...
 // Start defining your routes here
 app.get("/", (req, res) => {
-  res.send("Hello Technigo!");
+  res.send(listEndpoints(app));
```
CamillaHallberg, line 60: Nice that you are listing all of the endpoints at the home page. Endpoints naming is also good and easily understood :)

server.js
```diff
@@ -20,15 +15,70 @@ mongoose.Promise = Promise;
...
+
+})
+
+app.get("/streams/year/:release_year", async (req, res) => {
+  
+  const releaseYear = await Stream.find({release_year: req.params.release_year})
+  res.send(releaseYear)
+
+})
+
```
CamillaHallberg, line 81: Code looks nice and is easy to read. Well done :)

CamillaHallberg (approved):
Hi Ida!

Great job on this first week of using MongoDB! :) Your code is clean, easy to read and your naming of variables are good. Going through the requirements for this week I can see that you fulfill the blue level by having at least 2 routes, you have used the Mongoose model for the data and your API is RESTful! :) The only thing that has caught my attention is that when I'm on the homepage I can see that you have added an endpoint called "type" but I cannot see any code for this.

All in all, great job with Mongo this week and I'm looking forward to teaming up with you next with for the pair project! :)

PR #302, opened 2022-05-15
Completed the code

NabeelMansour:


server.js
```diff
@@ -2,34 +2,101 @@ import express from "express";
...
+      error: "Not found",
+      success: false,
+    });
+  } else {
+    res.status(200).json({
+      data: songByGenre,
+      success: true,
+    });
+  }
+});
```
michaelchangdk, line 96: I could get this part to work with the genre "pop" - but it would be nice to get an endpoint that sent all the data, so I could see what I could search for!

server.js
```diff
@@ -2,34 +2,101 @@ import express from "express";
...
+    trackName: trackName,
+  });
+  res.send(singleSong);
+ } else {
+  const singleSong = await Song.find({
+    artistName: artistName,
+  });
+  res.send(singleSong);
+}
+});
```
michaelchangdk, line 73: I think the reason why this endpoint isn't working is because it's got the same path as the route below. A little hard to test - I don't know much about req.query, I've only been using req.params.

michaelchangdk (commented):
Wish there was an endpoint to retrieve all the songs, so I could test better. But the genre param did work :)

PR #303, opened 2022-05-16
Mongo backend API Lovisa

LovisaBrorson:
https://mongo-api-lovisa.herokuapp.com/netflixshows

### DIFF
--- a/server.js
+++ b/server.js
@@ -59,21 +59,21 @@ app.get("/netflixshows/director/:director", async (req, res) => {
   const singleDirector = await Netflixshow.findOne({
     director: req.params.director,
   })
-  res.send(singleDirector)
+  res.status(200).send(singleDirector)
 })
 
 app.get("/netflixshows/title/:title", async (req, res) => {
   const netflixTitle = await Netflixshow.find({
     title: req.params.title,
   })
-  res.send(netflixTitle)
+  res.status(200).send(netflixTitle)
 })
 
 app.get("/netflixshows/showsbytype/:type", async (req, res) => {
   const showtype = await Netflixshow.find({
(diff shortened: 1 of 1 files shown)


server.js
```diff
@@ -1,35 +1,105 @@
...
 // If you're using one of our datasets, uncomment the appropriate import below
 // to get started!
 // import avocadoSalesData from "./data/avocado-sales.json";
 // import booksData from "./data/books.json";
-// import goldenGlobesData from "./data/golden-globes.json";
-// import netflixData from "./data/netflix-titles.json";
+// import goldenGlobesData from "./data/golden-globes.json"
+import netflixShows from "./data/netflix-titles.json"
+//import { all } from "express/lib/application"
 // import topMusicData from "./data/top-music.json";
```
kolkri, line 13: You should remove the imported data that are commented out, because you are not using them.

server.js
```diff
@@ -1,35 +1,105 @@
...
+  res.send("Hello! This is the API for week 18!!")
+})
+
+app.get("/netflixshows/director/:director", async (req, res) => {
+  const singleDirector = await Netflixshow.findOne({
+    director: req.params.director,
+  })
+  res.status(200).send(singleDirector)
+})
+
```
kolkri, line 64: You could consider adding try and catching error message / so you will have better backend because you get error messages and know what is wrong if is is not finding a director with given value

kolkri: You can added this to every endpoints

kolkri (commented):
Hi Lovia, your herokuapp link didn't work for me, did you connect that to your database in MongoDB atlas. That could be one reason why it is not working.

PR #304, opened 2022-05-16
project-mongo-api JaEngd

JaEngd:


server.js
```diff
@@ -24,12 +26,87 @@ const app = express();
...
+  const booksByMinPages = await Book.find().sort({num_pages:+1}).limit(1)
+  res.send(booksByMinPages[0])
+})
+
+// Sorts out the book with the least amout of pages.
+app.get("/booksData/average_rating", async (req, res) => {
+  const booksByRating = await Book.find().sort( { average_rating:[-1] } )
+  res.send(booksByRating)
+})
+
```
savannah-hayes, line 108: Unfortunately, I can't seem to get any data back from your deployed Heroku link. The API doesn't seem to work when I put in the endpoints you setup. I am wondering if it is working on your side locally because I don't seem to see anything from your deployed link. 

Did you add the environmental variables in a .env file of your project and add the environmental variables in Heroku?

server.js
```diff
@@ -24,12 +26,87 @@ const app = express();
...
+      newBook.save();
+    });
+  };
+  seedDatabase();
+}
+
 app.get("/", (req, res) => {
-  res.send("Hello Technigo!");
+  res.send(listEndpoints(app));
+});
```
savannah-hayes, line 55: It is great that you listed all the endpoints in the root path

savannah-hayes (commented):
Your code seems to look good, unfortunately I can't add more than that because I couldn't test your endpoints. Your endpoints are probably setup right but I can't seem to see them on my end, I mentioned in my comments that maybe you can see them locally on your computer.

I would be happy to help you setup your environmental variables if they haven't been setup. :)

PR #305, opened 2022-05-17
Kristiina's project-mongo-api

kolkri:


server.js
```diff
@@ -24,12 +19,143 @@ const app = express();
...
+  const { bookID } = req.params;
+  console.log(req)
+  const bookById = booksData.find((item) => item.bookID === +bookID);
+
+  if (!bookById) {
+    res.status(404).send('No book found!')
+  };
+  
+  res.json(bookById);
+})
```
tiiliu, line 138: I wasn't able to return a book using this bookId and I noticed that the error message it was giving me was the error message that you have for the other id endpoint, on row 76. The paths seem to be identical so perhaps that it causing the problem? I also noticed that you are mapping trough the booksData which you are importing from your json-file, so this data is not coming from the database?

server.js
```diff
@@ -24,12 +19,143 @@ const app = express();
...
+    res.status(404).json({
+      response: 'No book found with that name',
+      success: false
+    }) 
+  } else {
+    res.status(200).json({
+      response: bookByName,
+      success: true
+    }) 
+  }
```
tiiliu, line 155: I was unfortunately not able to fetch a specific book with the title either. I notice that you're mapping through the json data here too. You seem to have the data in the database because returning all the books with the main query and fetching the id with the mongoose id works, so perhaps there's something not working because of that? I think you could use the mongoose query methods for fetching this data from the database?

You could do this by fetching the book by name from the database using regex, something like this for example:

`const { bookName } = req.params
const bookByName = await Book.find({
title: new Regexp(bookName, "i")
})`

Or you could change your const on row 62 to a let and then do if statements inside your main query and then assign your let booksToShow to be your result of the query according to the if statements. Something like this:

`let booksToShow = await Book.find(req.query
if (req.query.bookName) {
const bookByName = await Book.find({
title: new Regexp(req.query.bookName, "i")
})
booksToShow = bookByName`

It also gave me a "successful" request as the message from that came back was success: true even though I was not able to fetch a book. So perhaps also wrapping the whole request into a try catch form would help in spotting what went wrong with the request for the user too?  😊

server.js
```diff
@@ -24,12 +19,143 @@ const app = express();
...
+    response: booksToSend,
+    success:true
+  })
+  
+});
+
+// //get book data by id
+app.get('/books/id/:bookID', (req, res) => {
+  const { bookID } = req.params;
+  console.log(req)
```
tiiliu, line 130: Here's one forgotten console.log 😊

tiiliu (commented):
This week was a struggle with the deployment and everything, but you clearly have the data in the database and you are filling the blue level requirements, that's great! 🎊 😊  One thing I think would be good to add is documentation about the endpoints to the readme document for example, perhaps that is something that you just haven't had the chance to do yet. I left some comments inside the code, hope they can be helpful for you. 🤗

PR #307, opened 2022-05-18
Mongo db

sk-phan:


qwer-ctrl (commented):
Good work again Suki :) code is compact and fill requirements. I like you're average rating data collection, where you have use find and get filter both. Also deployed site works and showing endpoints correctly :)

PR #308, opened 2022-05-19
Project Mongo Api, Tiina Liukkonen

tiiliu:


server.js
```diff
@@ -1,35 +1,308 @@
...
+      "/tracks/genres/:genre":"Get tracks by genre (for example: /tracks/genres/latin)",
+      "/tracks/artists/:artist":"Get a track by the artists name (for example: /tracks/artists/ed sheeran)",
+      "/tracks/songtitles/:songtitle":"Get a specific track by the songtitle (for example: /tracks/songtitles/truth hurts)",
+      "/tracks/id/:id":"Get a track by _id (for example: /tracks/id/6283e58592aafedccbeb0e0e)",
+      "/tracks?genre=${genre}":"Get tracks by genre (for example: /tracks?genre=hip hop)",
+      "/tracks?artist=${artist}":"Get all tracks from a specific artist (for example: /tracks?artist=billie eilish)",
+      "/tracks?songtitle=${songtitle}":"Get a specific track by the songtitle (for example: /tracks?songtitle=shallow)"
+    }]
+  })
+})
```
Mimmisen, line 85: You are showing all your endpoints very clearly and with descriptions, that is very nice!
It can be a lot to write, you could try out using the npm package express list endpoints. It shows all the endpoints that you create automatically. Then you only need to document in the readme what endpoints you have!

server.js
```diff
@@ -1,35 +1,308 @@
...
+    })
+  }
+})
+
+
+// get a single track based on database id
+app.get("/tracks/id/:id", async (req, res) => {
+  try {
+    const trackById = await Track.findById(req.params.id)
+    if (trackById) {
```
Mimmisen, line 267: In last weeks mongo api common pitfalls doc they want us to write endpoints like this:
_When you are accessing a single document with the key, use syntax like this: /shows/:id
rather than /shows/id/:id_

server.js
```diff
@@ -1,35 +1,308 @@
...
+app.get("/tracks/genres/:genre", async (req, res) => {
+  try {
+    const { genre } = req.params
+    const tracksByGenre = await Track.find({
+      genre: new RegExp(genre, "i")
+    })
+    if (tracksByGenre) {
+      res.status(200).json({
+        data: tracksByGenre,
+        success: true
```
Mimmisen, line 192: I can't make the endpoints with the query work, genre, artist, and songtitle...
I tested them out in your previous week's project and there they worked. 
From the live session on Wednesday Daniel goes through this, maybe take a look at that video and see if you can figure out the problem :)

Mimmisen (approved):
Great work this week Tiina! 🥇 💯 
I'm so impressed that you have created a lot of endpoints! They are very useful and it's nice that you implemented the random one! You have great documentation for the endpoints and I really like that you show how to use them! You also have nice error handling :) 
An awesome job! 👍

PR #314, opened 2022-11-25
Project-mongo-API Kaja Wilbik

KajaWilbik:


server.js
```diff
@@ -14,10 +16,51 @@ const mongoUrl = process.env.MONGO_URL || "mongodb://localhost/project-mongo";
...
+      category: "best actor",
+      nominee: "Johny Deep",
+      win: true
+    
+    });
+    addedNominate.save();// we need to put it in the reset function because if is after the deleteMany happens before and we do not update the DB
+    addedNominate2.save()
+  }
+  resetDataBase();
+}
```
SofieRydmark, line 58: right now you only have two data objects in your api, and to add more from the golden globe seperately would be too much :) a suggestion to get more data in to your DB at the same time from the json file is you can write a forEach iteration:
```

if(process.env.RESET_DB) {
  const resetDataBase = async () => {
    await Nomination.deleteMany();
    goldenGlobesData.forEach(singleNomination => {
      const newNomination = new Nomination(singleNomination);
      newNomination.save();
    })
  }
  resetDataBase();
}
```

server.js
```diff
@@ -26,10 +69,58 @@ app.use(express.json());
...
+        data: "There is not data with this category. Try: <br> best actor <br> best tragic comedy "
+      })
+    }
+  }
+     catch (err) {
+      res.status(400).json({
+        success: false,
+        status_code: 400,
+        error: "Invalid category!! Try: `best actor`, `best tragic comedy`"
+      })
```
SofieRydmark, line 111: not sure why i can't see the error messages when entering an invalid category, i only see status code 200 and an empty array []. I think if you change "data" to "error" on line 102 instead it will show the text also? :)

SofieRydmark (approved):
good job! You already commented that you want to add more endpoints and data later so i don't need to write as well. maybe take a look at the error message since that is not coming up right now :)

PR #315, opened 2022-11-25
Project-MongoApi - David Ballester

Fonnt:
https://project-mongo-api-5ympbqoweq-lz.a.run.app/


server.js
```diff
@@ -26,10 +56,238 @@ app.use(express.json());
 
 // Start defining your routes here
 app.get("/", (req, res) => {
-  res.send("Hello Technigo!");
+  res.status(200).json({
+    Message: "Top-Music data",
+    data: listEndpoints(app)
+  });
+});
+
```
juliaostedt, line 64: Would suggest having a look at your routes order and see if perhaps /lists and /songsquery should be moved further up in your code? There was a section on this in week 13 common pitfalls, here's a screen shot. 
image

juliaostedt (commented):
Well done! Your code is easy to follow and your comments are clear and helpful along the way. Perhaps you could update your model and remove what's not used, but I'm not sure if that's best practice or if it's better to have them all written out for future reference... 

Your error messages works well and I think it's great that you're using the regular expressions! 

Small reminder to update your read-me file. :)

PR #321, opened 2022-11-25
Linda Malm - MongoDB API

malmen237:
Link to the API:

https://project-mongo-api-pi3gkxc3ea-lz.a.run.app/


README.md
```diff
@@ -1,13 +1,23 @@
...
-## View it live
+https://project-mongo-api-pi3gkxc3ea-lz.a.run.app/
 
-Every project should be deployed somewhere. Be sure to include the link to the deployed project so that the viewer can click around and see what it's all about.
+
+## Frontend
+
+### View it live
+
+https://nes-games-mongodb.netlify.app/
```
jessikalind, line 23: Very nice touch building a frontend! The visualisation of the content of the API is very helpful when trying to grasp what we are actually building in the backend!

backend/README.md
```diff
@@ -0,0 +1,14 @@
+# Project Mongo API
+
+GET
+
+#### HOME:                 /
+#### GAMESLIST:            /games
```
jessikalind, line 6: Great that you use plural (gameS) when creating this endpoint!

backend/server.js
```diff
@@ -0,0 +1,94 @@
...
+app.use(cors());
+app.use(express.json());
+
+app.use((req, res, next) => {
+  if (mongoose.connection.readyState === 1) {
+    next()
+  } else {
+    res.status(503).json({ error: "Service unavaliable" })
+  }
+});
```
jessikalind, line 18: Super that you added an error msg for the 503.

frontend/src/components/Details.js
```diff
@@ -0,0 +1,23 @@
...
+
+const Details = () => {
+  const [listItem, setListItem] = useState([])
+  const { gameId } = useParams();
+
+  useEffect(() => {
+    fetch(`https://project-mongo-api-pi3gkxc3ea-lz.a.run.app/games/${gameId}`)
+      .then((data) => data.json())
+      .then((json) => setListItem(json.body))
+  }, [gameId])
```
jessikalind, line 12: It really is helpful seeing it in action!

backend/server.js
```diff
@@ -0,0 +1,94 @@
...
+      success: true,
+      nesGames: response
+    })
+  } catch (err) {
+    res.status(400).json({ 
+      success: false,
+      error: "Invalid id-request"
+    })
+  }
+});
```
jessikalind, line 67: I'm not at all sure I'm right, but perhaps the error msg could be altered a bit, since the id is not directly involved here. 

Also, I'm wondering if it could be possible to add some alternative error msg, since an empty array is returned with success "true", if I for example misspell a developer (using the endpoint /games?developer=*developer*). Perhaps, something like if (response.body === 0) { sucess: false etc.

jessikalind (approved):
Linda! 🥳 This is possibly my shortest code review ever - and that is simply because I don't really have anything to say but: great job! The API works well, and it is great that you managed to build a frontend as well! It's super to be able to see it all visually.

The only remark I have is concerning one of the error-msges (see comment for more detail).

PR #326, opened 2022-11-27
Jessika's MongoDB project

jessikalind:
https://project-mongo-api-rnx2qpxu2a-lz.a.run.app/

### DIFF
--- a/server.js
+++ b/server.js
@@ -3,7 +3,9 @@ import cors from "cors";
 import mongoose from "mongoose";
 // import topMusicData from "./data/top-music.json"
 
-const mongoUrl = "mongodb+srv://user:Jx7Uw4k7yieA1laO@cluster0.whdukaj.mongodb.net/project-mongo?retryWrites=true&w=majority";
+const mongoUrl = process.env.MONGO_URL || "mongodb://localhost/project-mongo";
+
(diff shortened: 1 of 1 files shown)


server.js
```diff
@@ -2,34 +2,132 @@ import express from "express";
...
 app.use(cors());
 app.use(express.json());
 
-// Start defining your routes here
+//ROUTES
+// Main route
 app.get("/", (req, res) => {
-  res.send("Hello Technigo!");
+  res.json({ 
+    responseMessage: "Welcome to Jessika's Spotify API 🎶!",
```
CeciliaMichele, line 52: Nice starting page with all the different routes

server.js
```diff
@@ -2,34 +2,132 @@ import express from "express";
...
+app.get("/songs/", async (req, res) => {
+  
+  const {genre, trackName, artistName} = req.query
+  const response = {
+    sucess: true,
+    body: {}
+  }
+
+  const matchAllRegex = new RegExp(".*")
+  const genreQuery = genre ? genre : {$regex: matchAllRegex, $options: 'i'}
```
CeciliaMichele, line 73: Fun that you used Regex!

CeciliaMichele (commented):
I think this site was really good especially since I know you struggled a bit last week, I wouldn't say this is the bare minimum at all, it's a lot of details to it and really fun that you tried out the Regex, genre query etc. Fun website! Really good job this week!

PR #329, opened 2022-11-27
Project - Mongo API - Antonella Cardozo

Sailornina:
- Documented version: https://project-mongo-api-oac2twtmjq-lz.a.run.app/api-docs/
- Google Cloud version: https://project-mongo-api-oac2twtmjq-lz.a.run.app/


server.js
```diff
@@ -1,32 +1,210 @@
...
+ * @openapi
+ * /endpoints:
+ *   get:
+ *     description: Get API endpoints.
+ *     responses: 
+ *       200: 
+ *         description: Successful response. 
+ */
+app.get("/endpoints", (req, res) => {
+  res.send(listEndpoints(app))
```
Petrasoderstrom1612, line 102: Nice that you also user the express for endpoints!

server.js
```diff
@@ -1,32 +1,210 @@
...
+ *     description: Get all World Cup.
+ *     responses: 
+ *       200: 
+ *         description: Successful response. 
+ */
+app.get("/cups", async (req, res) => {
+  let query = req.query || {}
+  try {
+    let cups = await Cup.find(query)
+    res.json(cups)
```
Petrasoderstrom1612, line 118: Here, I suggest to endorce the status 200 `res.status(200)json(cups)`

server.js
```diff
@@ -1,32 +1,210 @@
...
+ *         description: Get a World Champion with Matching ID.
+ *     responses: 
+ *       200: 
+ *         description: Successful response. 
+ */
+app.get("/cups/:_id", async (req, res) => {
+  try {
+    const cupId = await Cup.findById(req.params._id)
+    if (cupId) {
+      res.json(cupId)
```
Petrasoderstrom1612, line 143: same here with status 200 as mentionnned above

server.js
```diff
@@ -1,32 +1,210 @@
...
+ *         schema:
+ *           type: string
+ *         description: Get a World Champion with Matching ID.
+ *     responses: 
+ *       200: 
+ *         description: Successful response. 
+ */
+app.get("/cups/:_id", async (req, res) => {
+  try {
+    const cupId = await Cup.findById(req.params._id)
```
Petrasoderstrom1612, line 141: What you could do in this part is to destructurize the _id into a variable before try ( row 140) `const { id } = req.params `which would then improve to user experience as you avoid the _ character hence `const cupId = await Cup.findById(req.params.id)` and the route to "/cups/:id"

server.js
```diff
@@ -1,32 +1,210 @@
...
+    if (cupId) {
+      res.json(cupId)
+    } else {
+      res.status(404).json({ error: "Id not found" })
+    }
+  } catch (error) {
+    res.status(400).json({ error: "Invalid Id" })
```
Petrasoderstrom1612, line 148: Well done distinguishing between error 404 and 400 helping the front end part!

server.js
```diff
@@ -1,32 +1,210 @@
...
+ *           type: string
+ *         description: Get the result of a World Champion.
+ *     responses: 
+ *       200: 
+ *         description: Successful response. 
+ */
+app.get("/cups/winner/:winner", async (req, res) => {
+  const { winner } = req.params
+  try {
+    const winnerCup = await Cup.find({ winner: { $regex: new RegExp("^" + winner, "i") } })
```
Petrasoderstrom1612, line 170: Very well done with regex!

PR #330, opened 2022-11-27
Project Mongo API | Sofia Andersson

Sofia-Andersson:
https://project-mongo-api-ircjrh2jfq-lz.a.run.app/


server.js
```diff
@@ -1,35 +1,201 @@
...
+    });
+  }
+});
+
+
+// Display products by lowest price
+app.get("/products/lowestprice/", async (req, res) => {
+  try {
+    const products = await Product.find({}).sort({price:1});
+    if (products) {
```
MariaWellander, line 105: Cool to add the .sort() to sort by price in ascending order :)

MariaWellander (commented):
### Am I able to understand the code easily? Does the code have lots of duplicates? Could it be tidied up?

I'm able to understand the code easily - super job! What about tidying it up by removing some comments?

### Does it work or does anything look broken?

The base path works but then going to the other routes doesn't work for me - though the route idea looks super nice in the code!

### Does it follow the general and blue requirements?

Yes! What I can see it even meets up the requirements for the red level - wonderful!

### To sum it up

Very good work, Sofia! You should be very proud of yourself. Your first Mongo API and Mongoose model - wow! I really like it that you managed to add frontend too + that you are presenting the routes in the base path. And as mentioned, I really liked that you picked a dataset from Kaggle. Fun with the Swedish touch to it - IKEA :)

PR #331, opened 2022-11-27
Linneas Music project MongoDB API

LinneaWa:
https://project-mongo-api-jbma6srvnq-lz.a.run.app/


server.js
```diff
@@ -1,14 +1,9 @@
 import express from "express";
 import cors from "cors";
 import mongoose from "mongoose";
+import listEndpoints from "express-list-endpoints";
```
katirum, line 4: I really like this snippet of code because it makes it easier to go through your endpoints. Might do that next time!

server.js
```diff
@@ -17,6 +12,65 @@ mongoose.Promise = Promise;
...
+  valence: Number,
+  length: Number,
+  acousticness: Number,
+  speechiness: Number,
+  popularity: Number
+
+})
+// The reset of database and adding new songs or albums 
+// was great for educational purpose but is not to be used live.
+//
```
katirum, line 35: Good explanation of why you have the RESET_DB, it shows that you understand the purpose of it.

server.js
```diff
@@ -17,6 +12,65 @@ mongoose.Promise = Promise;
...
+      danceability: 45,
+      loudness: 90,
+      liveness: 80,
+      valence: 34,
+      length: 937,
+      acousticness: 23,
+      speechiness: 46,
+      popularity: 97
+     })
+    await testSong.save()
```
katirum, line 65: This is good as a practice, but you might not need this anymore since your endpoints work :)

server.js
```diff
@@ -26,7 +80,94 @@ app.use(express.json());
...
+  } catch (error) {
+    res.status(400).json({
+      success: false,
+      message: "Invalid id",
+      body: {
+        topMusicData: {}
+      }
+    });
+  }
+})
```
katirum, line 140: This endpoint works with findById, it shows the correct data. However, I could not get to error message : songs not found when I entered the wrong ID. It went straight to error 400 right away.

server.js
```diff
@@ -26,7 +80,94 @@ app.use(express.json());
...
+  } catch (error) {
+    res.status(400).json({
+      success: false,
+      message: "Invalid bmp",
+      body: {
+        topMusicData: {}
+      }
+    });
+  }
 });
```
katirum, line 171: This endpoint also works well to show the correct bpm, but the error did not show up when I entered bpm number that does not exist in the data. One way you can do is to use .length. So for example;

> if (songByBpm.length !== 0)

That way, it will only return data that has length more than zero and also only the ones that match the data. If there is no length of data that does not match, it will return your error message.

katirum (commented):
Hi Linnea, 

Nice work for this project. I really like your way of showing endpoints. All of them are working and shows all the data that I want. 
You followed the instructions and you also fulfilled the blue requirements.
There is just tiny things that you can modify regarding error messages. Other than that, SUPERB WORK! 🎉😎

reviewed by: Kartika

PR #332, opened 2022-11-27
Joel Öhman project-mongo-api Week 14 etc etc

joeohm:
https://project-mongo-api-h6oeu2xskq-uc.a.run.app/


server.js
```diff
@@ -24,9 +27,130 @@ const app = express();
...
+    await Song.deleteMany();
+    topMusicData.forEach((singleSong) => {
+      const newSong = new Song(singleSong);
+      newSong.save();
+    });
+    netflixData.forEach((singleMovie) => {
+      const newMovie = new Movie(singleMovie);
+      newMovie.save();
+    });
+  };
```
LinneaWa, line 73: Cool that you used two datasets for this project. Nice personal touch! I'm wondering how come only the song-collection get deleted when reseting database. Why doesn't the movies-collection keep adding on?

server.js
```diff
@@ -24,9 +27,130 @@ const app = express();
...
+          message: 'Song was not found'
+        }
+      });
+    }
+  } catch (error) {
+    res.status(400).json({
+      success: false,
+      body: {
+        message: 'Invalid id'
+      }
```
LinneaWa, line 115: Great job with the status and error-messages. I've tested them and they work like they should. 👏

LinneaWa (commented):
The backend is solid and work like a charm. Amazing job with the frontend, I'm really impressed! (and curious, will need to ask you a few questions 😅) 

Super great job this week! 👏

server.js
joeohm: Haha, nice catch! I suppose it would create a lot of copies of movies if I ran it several times with RESET_DB=true. Will try it out and fix.

PR #333, opened 2022-11-27
project-mongo-api, Neah Rockstroh

Archkrull:


NinaBerggren (commented):
Great job Neah!
You have absolutely made the blue requirements. Database is running and the endpionts work. Nicely written code with clear status massages. Over all structure is well planed and executed, nice that you have the starting route  "games" and then have built from there, makes it easy to try the endpoints.
Keep up the fantastic work =)
// Nina

PR #334, opened 2022-11-27
Project Mongo Api - Elin Segelöv

ElinSegelov:
API is live at https://project-mongo-api-sz4vw4o3jq-lz.a.run.app/


server.js
```diff
@@ -1,35 +1,132 @@
...
+app.use((req, res, next) => {
+  if (mongoose.connection.readyState === 1) {
+    next()
+  } else {
+    res.status(503).json({ error: 'Service unavailable' })
+  }
+})
+
+// endpoints 
 app.get("/", (req, res) => {
```
Sailornina, line 58: This endpoint is not working. However, I think that you can use npm listEndpoints if you want to list the endpoints.

server.js
```diff
@@ -1,35 +1,132 @@
...
+      body: {
+        message: "Invalid id"
+      }
+    })
+  }
+});
+
+app.get("/songs/genre/:genre", async (req, res) => {
+  try {
+    const genreCollection = await Song.find({genre: req.params.genre}).exec();
```
Sailornina, line 105: Using Regex in this endpoint would be fantastic because if the user types "Canadian pop" instead of "Canadian pop"  gets an empty array. It's case-insensitive.

Sailornina (commented):
Hi Elin! Good job! You were able to meet the goals for the week :D Happy coding!!

PR #337, opened 2022-11-27
Project Mongo API | Younas

YoTesfam:
https://project-mongo-api-sostxdjqaa-lz.a.run.app/

### DIFF
--- a/server.js
+++ b/server.js
@@ -49,6 +49,16 @@ if(process.env.RESET_DB){
 }
 const port = process.env.PORT || 8080;
 const app = express();
+
+app.use((req, res, next) => {
+  if (mongoose.connection.readyState === 1) {
+    next()
+  } else {
+    res.status(503).json({
+      error: `Service unavailable`
+    })
+  }
+})
 // Add middlewares to enable cors and json body parsing
 app.use(cors());
 app.use(express.json());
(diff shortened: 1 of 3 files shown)


server.js
```diff
@@ -1,35 +1,180 @@
...
+app.get("/songs/", async (req, res) => {
+
+  const {genre, danceability} = req.query;
+  const response = {
+    success: true,
+    body: {}
+  }
+  const matchAllRegex = new RegExp(".*");
+  const matchAllNumeric = new RegExp("[0-9]");
+  const genreQuery = genre ? genre : {$regex: matchAllRegex,  $options: 'i' };
```
Scarylie, line 138: this is different from Daniels code that looks like this ``` const genreQuery = genre ? genre : matchAllRegex;``` what are you thing to do? I dont understand.

server.js
```diff
@@ -1,35 +1,180 @@
...
+  const danceabilityQuery = danceability ? danceability : {$gt: 0, $lt: 100};
+
+  try {
+    // if ( req.params.genre && req.params.danceability) {
+      response.body = await Song.find({genre: genreQuery, danceability:danceabilityQuery}).limit(2).sort({energy: 1}).select({trackName: 1, artistName: 1})
```
Scarylie, line 143: I dont know why the query filtering is not working properly

server.js
```diff
@@ -1,35 +1,180 @@
...
+  //sorts all songs based on popularity rating
+  app.get("/songs/popularity", async (req, res) => {
+    try {
+      let popular = await Song.find()
+      popular = popular.sort((a, b) => b.popularity - a.popularity)
+      res.json({ length: popular.length, data: popular })
+    } catch {
+      res.status(400).json ({ error: `invalid request` })
+    }
+  })
```
Scarylie, line 90: I really like that you managed to sort the Song list. I tried to do it but couldn't get it to work so well done.

Scarylie (approved):
Well done! I think you have managed to complete the requirements. I like that you found different ways to solve the problems even though i don't understand it all. Could be nice with some comments when using things that are new.  Have a look at the queries since i don't think the filtering is working properly.

PR #339, opened 2022-11-27
Project Mongo API Carina Moreno

carimore:


server.js
```diff
@@ -17,6 +17,36 @@ mongoose.Promise = Promise;
 // Defines the port the app will run on. Defaults to 8080, but can be overridden
 // when starting the server. Example command to overwrite PORT env variable value:
 // PORT=9000 npm start
+
+const Song = mongoose.model("Song", {
```
jessnmo, line 21: could now update this in a schema! but looking good

PR #344, opened 2022-11-27
Week 14, Project MongoDB

miadahlgren:


server.js
```diff
@@ -26,10 +46,49 @@ app.use(express.json());
 
 // Start defining your routes here
 app.get("/", (req, res) => {
-  res.send("Hello Technigo!");
+  res.send("Hello! Try out the routes '/titles' or '/titles/id/:id' ");
+});
```
YoTesfam, line 50: I can see on your reference that you had "dotenv" but are not using it here.
I recommend it, I would add in this section:

app.get("/", (req, res) => {
res.send({ "Routes": listEndpoints(app)});
});

Dont forget to add this: dotenv.config()
somewhere below the imports

server.js
```diff
@@ -1,22 +1,42 @@
...
+    netflixData.forEach(item => {
+      const newTitle = new Title(item);
+      newTitle.save();
+    })
+  }
+  resetDataBase();
+}
+
 const port = process.env.PORT || 8080;
 const app = express();
```
YoTesfam, line 42: Another thing that I learned is the below code:
This is for when you click on your backend link and if it for exemple takes some time for it to connect or not.

//an express middleware to check if mongoose is ready and throw an error if not. Note: Should always be added before the router.
app.use((req, res, next) => {
  if (mongoose.connection.readyState === 1) {
    next()
  } else {
    res.status(503).json({
      error: `Service unavailable`
    })
  }
})

YoTesfam (commented):
Hi Mia,

I was missing your link to check if what you had did was working.
But Your code looked neat and clean, I understood it all.

I have added two recommendation that I like to use in my code.

Take care and good job!

PR #353, opened 2022-11-27
Project Mongo API | Julia Östedt

juliaostedt:
https://mongo-api-3zjzn6vrhq-lz.a.run.app/


server.js
```diff
@@ -1,35 +1,121 @@
...
+    Routes: [
+      {
+        "/goldenglobes": "Shows an array of all Golden Globes objects",
+        "/goldenglobes/year/:year_award": "Displays all nominees from a specific year",
+        "/goldenglobes/nominees/:nominee": "Displays information on a specific nominee",
+        "/goldenglobes/winners/:win": "Displays an array of all the winners or the ones who did not win. True/False",
+      },
+    ],
+  }
+  res.send(endpoints)
```
Kneedler, line 50: Nice solve with the const endpoints, i like that the user can get an over view!

server.js
```diff
@@ -1,35 +1,121 @@
...
+      });
+    }
+  } catch(error) {
+    res.status(400).json({
+      success: false,
+      body: {
+        message: "Invalid nominee name"
+      }
+    });
+  }
```
Kneedler, line 108: You can simplify this by destructuring the body.

`app.get("/globes/", async (req, res) => {
  const { nominee, ceremony } = req.query;
  const response = {
    success: true,
    body: {}
  }
  const nomineeQuery = nominee ? nominee : /.*/;
  const ceremonyQuery = ceremony ? ceremony : {$gt: 0, $lt: 100};
  try {
    response.body = await Globe.find({nominee: nomineeQuery, ceremony: ceremonyQuery})
    res.status(200).json({
      body: response
    });
  } catch (error) {
    res.status(400).json({
      success: false,
      body: { 
        message: error
      }
    })
  }
});`

As u mentioned about If Else, mine did not work last week, but this week with this code it worked better at least😅

Kneedler (commented):
Good job Julia! The endpoints work, it's easy to follow the code and it looks neat. I left a tip on how to simplify a bit (its hard to read tho, because the format got messed up... )

Keep it up! :D

PR #355, opened 2022-11-27
Johanna Rosenholm-project-mongo-api

RosenholmJohanna:


server.js
```diff
@@ -1,35 +1,171 @@
...
+// put and delete entries in database
+  if(process.env.RESET_DB) {
+  const resetDataBase = async () => {
+    await Song.deleteMany();
+    topMusicData.forEach(singleSong => {
+      const newSong = new Song(singleSong);
+      newSong.save();
+    })
+  }
+  resetDataBase(); //DO NOT EVER DO THIS IN PRODUCTION ENVIROMENT!
```
KajaWilbik, line 52: Good idea with commenting the part that was educational but we should not use in the production enviroment

server.js
```diff
@@ -1,35 +1,171 @@
...
 // Add middlewares to enable cors and json body parsing
 app.use(cors());
 app.use(express.json());
 
-// Start defining your routes here
+//Start route show endpoints
 app.get("/", (req, res) => {
-  res.send("Hello Technigo!");
+  res.status(200).json ({
+    data:({ "Available Routes for Top-Music" : listEndpoints(app) })
```
KajaWilbik, line 66: express list endpoints? What is it?

KajaWilbik (commented):
Nice and clean code, easy to follow. 

I checked the code in Postman and I got an error with the https://project-mongo-api-o2w727wcdq-lz.a.run.app/allsongs 

I am getting an empty array without data, which made it impossible to check the next endpoint with ID. I tried to debug it but I cannot find what is the problem. Was it working before?

PR #356, opened 2022-11-28
Project-Mongo-API by Fanny Cardell

fannycardell:


server.js
```diff
@@ -1,35 +1,133 @@
...
-// Add middlewares to enable cors and json body parsing
 app.use(cors());
 app.use(express.json());
 
-// Start defining your routes here
+// Routes defined here
 app.get("/", (req, res) => {
-  res.send("Hello Technigo!");
+ res.json({ 
+  Message: "Welcome to Top 50 at Spotify! Here you can look at all songs or search for specific song, artist, genre, energy and danceability by typing in the following strings after the adress in the browser:",
```
jesshansson, line 48: Good and informative message at the first page 👍

server.js
```diff
@@ -1,35 +1,133 @@
...
+  routes: {
+    "/songs": "look for all songs availible here",
+    "/songs?trackName=*the name of the track*": "search for a specific track",
+    "/songs?artistName=*the artist's name*" : "search for songs by a specific artist",
+    "/songs?genre=*name of the genre*": "search for a specific genre", 
+    "/songs?energy=*number of enery*": "search for a energy level",
+    "/Songs?danceability=*number of danceability*": "search for danceability level",
```
jesshansson, line 55: Cool feature, to be able to search based on energy and danceability level! 😀

server.js
```diff
@@ -1,35 +1,133 @@
...
+  const response = {
+    success: true,
+    body: {}
+  }
+  const matchAllRegex = new RegExp(".*");
+  const trackQuery = trackName ? trackName : {$regex: matchAllRegex,  $options: 'i' }
+  const artistQuery = artistName ? artistName : {$regex: matchAllRegex,  $options: 'i' }
+  const genreQuery = genre ? genre : {$regex: matchAllRegex, $options: 'i'}
+  const energyQuery = energy ? energy : {$gt: 0, $lt: 100};
+  const danceabilityQuery = danceability ? danceability : {$gt: 0, $lt: 100};
```
jesshansson, line 76: Really nice feature, but might need some small fix to work!

jesshansson (commented):
Great job! Your code looks really neat and professional, with advanced methods, so you can definitely be proud over yourself! I think you've done a good job choosing routes, they definitely cover interesting features of the songs  😊 I also think your API is RESTful with good names for the endpoints.

Now when I tried to search on an energy level, it doesn't quite seem to work though (I got some songs with a lower level). I have to admit that I don't really understand the code you've used, is it a range from 0-100? I just used " $gte: 80" to list songs with a level of 80 or higher, maybe that can be enough? It would be really cool if you could search on a level (instead of just listing them, as I do), but if so it could be good to write the span you can search for on the first page 👍 

Please ask if I don't make any sense! 😄

PR #367, opened 2023-05-11
Project mongoDB by Leo Thunell

khleoth:
https://project-mongo-api-3oe2ikwwwa-lz.a.run.app

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,13 +1,11 @@
 # Project Mongo API
 
-Replace this readme with your own information about your project.
-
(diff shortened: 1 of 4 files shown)


noraterbocs (commented):
I like it that you took the time and used more logic on validating the params or constructing the category endpoint for example. 
Although, I don't see the point to have these separate datasets (songs and movies) in the project unless you tie them together in an endpoint, but for practicing purposes it's ok. 
Good job, Leo!

PR #372, opened 2023-05-12
https://hello-touaoaeitq-lz.a.run.app/

EmiliaGA:


server.js
```diff
@@ -1,35 +1,88 @@
...
+app.get("/titles", async (req, res) => {
+  const allTheTitles = await Title.find({})
+  res.status(200).json ({
+    success: true,
+    body: allTheTitles
+  })
+})
+
+app.get("/titles/:id", async (req, res) => {
+  try {
```
smirrebinx, line 59: Good job with the try and catch. I will add this too in my code :-)

smirrebinx (commented):
Good job! It seems like your code does what it should and that it meets the requirements, however when I try to add the endpoint /titles in you deployed site it return _Service unavailable._ I'm not sure if it has to do with the connection to Mongo DB or it's an issue with Google Cloud?

PR #380, opened 2023-05-14
Malins Project-Mongo-API

MalinSkill:
I have a problem with the API not working straight away, only after you refresh like 2-3 times. But once it's up and running everything seems to work. I need to figure out what is causing this problem.


Cheroptera (commented):
Well done Malin! 👏 It's so fun to see how you always find smart solutions to use in your code and how interested you are in solving things 😊  You picked an interesting API to work with and got everything (well, something happened with the /companies/:id endpoint, but that could also be temporary) to work.

PR #391, opened 2023-05-14
Project Mongo API - Frida M. G. 

Cheroptera:
https://project-mongo-api-kzsug7gh3a-lz.a.run.app/


TessAquilon (commented):
Hi Frida!

Sorry about my late code review this week :(

The code is perfect, from what I can see, and all the endpoints are working. Nice to have quite a few different ones to choose from. Well done! There were some parts that I haven't used for myself yet and thus have very little to say about, for example the RESET_DB part and the "i". Also, very nice comments throughout the whole code.

Keep up the great work, we're almost done! xx

PR #397, opened 2023-05-14
Mongo API w.14 jl

JohLeo:


Nahnahke (approved):
Would like to start off by saying that you did a great job! I have nothing to complain about. Nice, clean code which is easy to follow and understand.

You might have been able to push yourself to do a couple of more endpoints, but really not necessary! ;)

Way to go, you're a great code Johanna!

PR #398, opened 2023-05-14
Week 14 - Alexander (incomplete)

Alexander-Gabor:


server.js
```diff
@@ -28,6 +149,61 @@ app.use(express.json());
...
+    popularity: Number
+})
+
+const Song = mongoose.model("Song", songSchema);
+
+
+
+app.get("/songs/id/:id", async (req, res) => {
+  try {
+    const singleSong = await Song.findById(req.params.id);
```
JosefinRobertsson, line 184: For an additional endpoint that gets all of whatever property you would like from your Schema, you could use Song.find instead of findById

server.js
```diff
@@ -28,6 +149,61 @@ app.use(express.json());
...
+      })
+    } else {
+      res.status(404).json({
+        success: false,
+        body: {
+          message: "Song not found"
+        }
+      })
+    }
+  } catch(e) {
```
JosefinRobertsson, line 198: I like to use 'err' instead of 'e' to make it very clear that it stands for error and not event (since event in JS is often shortened to a single e as well). The context makes it clear what you mean here and it is fully correct, it's just a personal preference of mine to add clarity

JosefinRobertsson (commented):
You’ve got most of the work done already; you got the schema and model and a route for fetching a single item from your database. It would be rather straightforward to create another route to handle get requests for a list of all songs (their names or whatever you choose). That way you would fulfill the requirements of the project with at least two endpoints and their specifications. If you felt like it you could add extra queries, but maybe just start with a simple list of all the tracknames so you can check off this project. I see you have done a lot of work to learn about this in the commented out code, and I am 100% sure you could finish this project quite easily. Good start and keep it up! You got this!

PR #399, opened 2023-05-14
Mongo API Songs - Daniel

dannebrob:
https://project-mongo-api-codalong-brbqsbv2fq-uc.a.run.app/

### DIFF
--- a/server.js
+++ b/server.js
@@ -1,6 +1,8 @@
-import express from "express";
-import cors from "cors";
-import mongoose from "mongoose";
+import express from 'express';
+import cors from 'cors';
+import mongoose from 'mongoose';
+import musicData from './data/top-music.json';
+import listEndpoints from 'express-list-endpoints';
 
 // If you're using one of our datasets, uncomment the appropriate import below
 // to get started!
@@ -10,7 +12,7 @@ import mongoose from "mongoose";
 // import netflixData from "./data/netflix-titles.json";
 // import topMusicData from "./data/top-music.json";
 
-const mongoUrl = process.env.MONGO_URL || "mongodb://localhost/project-mongo";
+const mongoUrl = process.env.MONGO_URL || 'mongodb://127.0.0.1/musicDB';
 mongoose.connect(mongoUrl, { useNewUrlParser: true, useUnifiedTopology: true });
 mongoose.Promise = Promise;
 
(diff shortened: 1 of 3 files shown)


server.js
```diff
@@ -24,9 +29,113 @@ const app = express();
...
+    musicData.forEach((songData) => {
+      const song = new Song(songData);
+      console.log(song);
+      song.save();
+    });
+  };
+  seedDB();
+}
+// Check that the database is running
+app.use((req, res, next) => {
```
MalinSkill, line 64: Nice to have this check!

server.js
```diff
@@ -24,9 +29,113 @@ const app = express();
...
+  if (mongoose.connection.readyState === 1) {
+    next();
+  } else {
+    res.status(503).json({ error: 'Service unavailable' });
+  }
+});
+
+// Root endpoint (with list of endpoints)
+app.get('/', (req, res) => {
```
MalinSkill, line 73: I like this one too. Easy way to get the overview

server.js
```diff
@@ -24,9 +29,113 @@ const app = express();
...
+  res.json(listEndpoints(app));
+});
+
+// Endpoint to get all songs
+app.get('/songs', async (req, res) => {
```
MalinSkill, line 78: great with the try and catch error

server.js
```diff
@@ -24,9 +29,113 @@ const app = express();
...
+      body: {
+        track: singletrack
+      }
+    });
+  } else {
+    res.status(404).json({
+      success: false,
+      message: 'Track not found',
+      body: {}
+    });
```
MalinSkill, line 115: Nice structure of the response.

server.js
```diff
@@ -24,9 +29,113 @@ const app = express();
...
+    res.json(songs);
+  } else {
+    res.status(404).json({ error: 'BPM not found' });
+  }
+});
+
+// endpoint to get songs by genre
+app.get('/genres/:genre', async (req, res) => {
+  const { genre } = req.params;
+  const songs = await Song.find({ genre: genre });
```
MalinSkill, line 133: Daniel showed us a way to improve this search, as it is now when you search for example pop you will only receive the exact matching category. Since there are many different pop categories, maybe you would like to look into this solution:
Create a case insensitive regex:
`const genreRegex = new RegExp(genre, 'i');`
When using this it will include more matches:
`const songs = await Song.find({ genre: genreRegex })`

MalinSkill (approved):
Nice work Daniel, good choice of endpoints and it works well.

PR #401, opened 2023-05-14
working version - Ylva 

YlvaKarlsson:
So many problems with the backend this week - don't even know "how to get this" - will book a 1:1 this week and hopefully someone can give me a thorough walk-through on how to get all this and how to remember it... Over and out from me :P


server.js
```diff
@@ -1,19 +1,56 @@
...
-// to get started!
-// import avocadoSalesData from "./data/avocado-sales.json";
-// import booksData from "./data/books.json";
-// import goldenGlobesData from "./data/golden-globes.json";
-// import netflixData from "./data/netflix-titles.json";
-// import topMusicData from "./data/top-music.json";
-
-const mongoUrl = process.env.MONGO_URL || "mongodb://localhost/project-mongo";
+//Should I use docker for below issue or what do I need to do to solve it?
+//I couldn't make the MONGO_URL work with the localhost-version - when I did that it stopped working locally
```
EmmaHoltegaard, line 8: The way I understand it, the MONGO_URL is for deployment, whereas "mongodb://localhost/project-mongo" is for what we can use, to see our work in Compass as we're developing it. So no need to change the link that was already there (álthought for some of us, "mongodb://127.0.0.1:27017/project-mongo" works better, for some reason)

server.js
```diff
@@ -1,19 +1,56 @@
...
+const mongoUrl = process.env.MONGO_URL || "mongodb+srv://ylva87:hHwxpsqpR3V62DS9@cluster0.hsexulf.mongodb.net/?retryWrites=true&w=majority";
```
EmmaHoltegaard, line 9: The link you've added here goes in the environment variable, when you deploy on Google Cloud. It should not be in your codebase at all.

server.js
```diff
@@ -1,19 +1,56 @@
...
+
+const seedDatabase = async () => {
+  const data = topMusicData
+  Song.collection.insertMany(data, function(err, r) {
+  })
+}
+// seedDatabase()
+
+//switch to cover error-types and generate response
+const handleErrors = (err, res) => {
```
EmmaHoltegaard, line 39: This is the first time I've seen error handling in a function for itself. Maybe it would be more simple to just write them in the routes?

server.js
```diff
@@ -1,19 +1,56 @@
...
-
-const mongoUrl = process.env.MONGO_URL || "mongodb://localhost/project-mongo";
+//Should I use docker for below issue or what do I need to do to solve it?
+//I couldn't make the MONGO_URL work with the localhost-version - when I did that it stopped working locally
+const mongoUrl = process.env.MONGO_URL || "mongodb+srv://ylva87:hHwxpsqpR3V62DS9@cluster0.hsexulf.mongodb.net/?retryWrites=true&w=majority";
 mongoose.connect(mongoUrl, { useNewUrlParser: true, useUnifiedTopology: true });
 mongoose.Promise = Promise;
 
+
+const Song = mongoose.model('Song', {
```
EmmaHoltegaard, line 14: You model seems to have everything it should :)

EmmaHoltegaard (commented):
I read the schedule wrong, and accidentally did a code-review for you last week as well. Oh well, you just get two reviews from me. I know you had an intense week, and didn't have much time for this project. Yet you've managed to write quite a lot. I've pointed out a few things, for when you return to this project.

server.js
YlvaKarlsson: Thanks Emma - Yeah I noticed that too :D

server.js
YlvaKarlsson: Probably! I got some help from a more experienced developer but we didn't manage to make it work anyhow. So I changed it to be in the routes instaead and it worked much better.

PR #403, opened 2023-05-15
Changes to be able to deploy

fikamate:


server.js
```diff
@@ -1,35 +1,86 @@
...
+const port = process.env.PORT || 9000;
 const app = express();
 
 // Add middlewares to enable cors and json body parsing
 app.use(cors());
 app.use(express.json());
 
+const { Schema } = mongoose;
+
+const movieSchema = new mongoose.Schema({
```
FionaKlacar, line 23: Is there a difference between doing this and just `const movieScheme = new Schema`?

server.js
```diff
@@ -1,35 +1,86 @@
...
+  show_id: Number,
+  title: String,
+  director: String,
+  cast: [String],
```
FionaKlacar, line 27: Interesting to see the square brackets; I guess because it will return a whole array here?

server.js
```diff
@@ -1,35 +1,86 @@
...
+    })
+  }
+  resetDatabase();
+}
 // Start defining your routes here
-app.get("/", (req, res) => {
-  res.send("Hello Technigo!");
+
+app.get("/", (req, res) => {  // this is the root route
+res.send(listEndpoints(app));
```
FionaKlacar, line 54: It looks from your endpoints that you only created one endpoint, called ‘movies’, but that endpoint doesn’t show all the movies, without the use of filtering the data down using queries.

server.js
```diff
@@ -1,35 +1,86 @@
...
+});
+
+// try this in postman by using the url http://localhost:9000/movies and then add the query parameters you want to search for like for example http://localhost:9000/movies?title=The%20Matrix or http://localhost:9000/movies?director=Quentin%20Tarantino
+app.get("/movies", async (req, res) => {
+  const { title, director, cast } = req.query;
+  const response = {
+    success: true,
+    body: {}
+  }
+  const titleRegex = new RegExp(title, "i"); //this is a regular expression that will search for the title in the database, the "i" means that it will be case insensitive
```
FionaKlacar, line 64: Cool idea to use RegExp!

server.js
```diff
@@ -1,35 +1,86 @@
...
+  const directorRegex = new RegExp(director, "i");
+  const castRegex = new RegExp(cast, "i");
+
+  try {
+    const searchResultFromDB = await Movie.find({ title: titleRegex, director: directorRegex, cast: castRegex })
```
FionaKlacar, line 69: What is the significance of doing `const searchResultFromDB`? Curious because I didn't think to do that!

server.js
```diff
@@ -1,35 +1,86 @@
...
+    if (searchResultFromDB) {
+      response.body = searchResultFromDB
+      res.status(200).json(response)
+    } else {
+      response.success = false,
+        res.status(500).json(response)
```
FionaKlacar, line 75: Did you mean to use the same status error here as below, and not, for example, a 404?

FionaKlacar (commented):
Cool to see your work, Annika! When the deployed site is up I'll be able to check out the endpoints that you made as well!

PR #405, opened 2023-05-15
Week 14: project-mongo-api (incomplete)

sammyolsson:
Still having some problems.


README.md
```diff
@@ -1,4 +1,4 @@
-# Project Mongo API
+# Project Mongo API (incomplete)
 
 Replace this readme with your own information about your project.
```
mvfrid, line 3: Write a little something about the task, and then also about what endpoints the user can expect :)

server.js
```diff
@@ -24,12 +17,141 @@ const app = express();
...
+  res.send({
+    Message: "Here you can browse the top music data. Below you find the endpoints you can use 🎶",
+    Routes: [
+      {
+        "/songs": "= all songs",
+        "/songs/id/:id": "= a specific song based on id, write the id of the song instead of :id",
+        "/songs/artist/:artistname": "= get the titles from a specific artist, write the artist instead of :artistname"
+      }
+    ],
+  })
```
mvfrid, line 69: Excellent way of showing the user what they can do with this API. Nice job!

server.js
```diff
@@ -24,12 +17,141 @@ const app = express();
...
 });
 
-// Start the server
+
+
+
+// ROUTE: ALL SONGS
+
+app.get("/songs", async (req, res) => {
+  const songs = topMusicData;
```
mvfrid, line 78: Now you are using the data directly from the local file, where I actually don't think you need the async function? If you want to fill the endpoint with Songs from your database, I would modify the code something like this:

```suggestion
  const songs = await Song.find();
  if (songs) ...
```

server.js
```diff
@@ -24,12 +17,141 @@ const app = express();
...
+      body: {}
+    })
+  }
+})
+
+// ROUTE: ARTIST
+
+app.get("/songs/artist/:artistname", async(req,res) => {
+  try{
+    const artistName = await Song.findOne({artistName: new RegExp(req.params.artistname,"i")});
```
mvfrid, line 100: Nice use of Regex!

server.js
```diff
@@ -24,12 +17,141 @@ const app = express();
...
+    if(artistName){
+      res.status(200).json({
+        success: true,
+        body: artistName
+      })
+    } else {
+      res.status(404).json({
+        success: false,
+        body: {
+          message: "ERROR! Artist not found!"
```
mvfrid, line 110: You could experiment with having the artist the user searched for as a parameter in the message :)

server.js
```diff
@@ -24,12 +17,141 @@ const app = express();
...
+// ROUTE: ALL SONGS
+
+app.get("/songs", async (req, res) => {
+  const songs = topMusicData;
+  if (songs) {
+    res.status(200).json({
+      success: true,
+      message: "OK",
+      body: {
+        topMusicData: songs
```
mvfrid, line 84: Name it something other than topMusicData, it gets a bit confusing that it's called the same as the data json file. Maybe "content"?

server.js
```diff
@@ -24,12 +17,141 @@ const app = express();
...
+if (process.env.RESET_DB) {
+  const resetDatabase = async () => {
+    await Song.deleteMany();
+    topMusicData.forEach((singleSong) => {
+      const newSong = new Song(singleSong);
+      newSong.save()
+    })
+  }
+  resetDatabase();
+}
```
mvfrid, line 54: Nice work on using data seeding!

mvfrid (commented):
I couldn't test your API since I know you are having issues with deployment, but I went through the code anyway and made a little review. It seems to do what it should and is really nicely organised. You are using consistent endpoint set up and error handling, which is really great!

I found a few areas where you could improve a few things (added in separate comments), but nothing major. Let me know if you want me to review it again when it's deployed. 

Nice job this week! You did really good, even though it's not fully complete :)

PR #409, opened 2023-05-17
Week 14 | Mongo API

Amanda-Sjstrm:


server.js
```diff
@@ -19,14 +20,79 @@ mongoose.Promise = Promise;
...
+  }
+  }
+  })
+
+const Quote = mongoose.model("Quote", quoteSchema);
+
+if (process.env.RESET_DB) {
+  const resetDatabase = async () => {
+    await Quote.deleteMany();
+    topMusicData.forEach((singleQuote) => {
```
edv-rd, line 48: Should not be called topMusicData I guess :)

edv-rd (commented):
Clean, easy to read and does the job it looks like (couldn't find a link to a deployed version). Good job!

server.js
Amanda-Sjstrm: Haha maybe this is why it wasn’t really working for me this week 😅

PR #410, opened 2023-05-19
Project mongo api

camcron:
https://project-mongo-api-45a55p775a-lz.a.run.app/


SofiaGerdmar (approved):
Great job this week! Your server.js is clean and easy to follow and the endpoints work as intended. I like that you added so many different variables to search and using different ways to do so. Good use of RegEx as well. I don't have any changes to suggest, everything looks great!

PR #411, opened 2023-05-19
Week 14 mongo-api

SandraMadeleine:
https://project-mongo-api-e54soatzoa-lz.a.run.app/

### DIFF
--- a/server.js
+++ b/server.js
@@ -61,14 +61,15 @@ app.get("/", (request, response) => {
 // https://localhost:8080/netflix-titles
 // This gets a list of all titles 
 app.get('/netflix-titles', async (request, response) => {
+  console.log('Received GET request for /netflix-titles');
   const netflixTitles = await NetflixTitle.find();
   response.json(netflixTitles);
 });
 
-// https://localhost:8080/netflix.titles/
-// This gets a single title when adding its id-number after the last formard slash
+// https://localhost:8080/netflix.titles/ and then you add the unique id number for the series you want
+// This gets a single title when adding its id-number after the last forward slash
 app.get('/netflix-titles/:id', async (request, response) => {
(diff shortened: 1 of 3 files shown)


server.js
```diff
@@ -7,13 +7,42 @@ import mongoose from "mongoose";
...
+  rating: String,
+  duration: String,
+  listed_in: String,
+  description: String,
+  type: String
+});
+
+const NetflixTitle = mongoose.model('NetflixTitle', netflixTitleSchema);
+
+if (process.env.RESET_DB) {
```
majazimnoch, line 36: you can also make an endpoint that will be resetting the DB instead of writing in the terminal: 
```
app.get("/reset", async (req, res) => {
  try {
    await resetDatabase();
    res.json({
      success: true
    }) 
  } catch (e) {
    res.json({
      success: false,
    })
  }
})
```

server.js
```diff
@@ -25,8 +54,30 @@ app.use(cors());
 app.use(express.json());
 
 // Start defining your routes here
-app.get("/", (req, res) => {
-  res.send("Hello Technigo!");
+app.get("/", (request, response) => {
```
majazimnoch, line 57: did you install // npm install express-list-endpoints and added
const listEndpoints = require('express-list-endpoints')

this would list your endpoint automatically after adding:
`app.get("/", (req, res) => {
  res.json(listEndpoints(app));
});`

majazimnoch (commented):
Good job, Sandra! Clean code with good comments. 
I added some comments reg resetting the db and an npm package called express list endpoints, instead of writing them manually, which I guess you did(?)
Also, it would be nice to update the read me with the deployment link.
However, I now try to search for the titles by /netflix-titles and unfortunately, the result is empty so probably some problem with the db?

PR #432, opened 2023-11-25
Emmys mongo-api-project

EmmyLJ:
I really had a hard time understanding everything this week and I still cant figure out everything, so have a lot of problems with the author part. Would love to get some feedback/help on that part. :)

### DIFF
--- a/server.js
+++ b/server.js
@@ -1,6 +1,7 @@
 import express from "express";
 import cors from "cors";
 import mongoose from "mongoose";
+import "dotenv/config";
 
 // If you're using one of our datasets, uncomment the appropriate import below
 // to get started!
@@ -10,8 +11,18 @@ import booksData from "./data/books.json";
 // import netflixData from "./data/netflix-titles.json";
 // import topMusicData from "./data/top-music.json";
 
-const mongoUrl = process.env.MONGO_URL || "mongodb://localhost/books";
-mongoose.connect(mongoUrl, { useNewUrlParser: true, useUnifiedTopology: true });
+const mongoUrl = process.env.MONGO_URL || "mongodb://localhost:27017/books";
+const connectToMongo = () => {
+  mongoose
+    .connect(mongoUrl, { useNewUrlParser: true, useUnifiedTopology: true })
+    .then(() => console.log("Connected successfully"))
+    .catch((err) => {
+      console.error("Connection error:", err);
+      setTimeout(connectToMongo, 5000);
+    });
+};
+
+connectToMongo();
 mongoose.Promise = Promise;
 
 //Creates a mongoose model named "Books"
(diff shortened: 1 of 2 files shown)


server.js
```diff
@@ -1,32 +1,117 @@
...
+app.get("/books", (req, res) => {
+  try {
+    res.json(booksData);
+  } catch (error) {
+    res.status(500).json({ error: "Internal Server Error" });
+  }
+});
+
+// Define the Author schema
+const authorSchema = new mongoose.Schema({
```
FridaKristian, line 61: I'm not sure, but I think you might need to import { Schema } from mongoose before using this..

server.js
```diff
@@ -1,32 +1,117 @@
...
+});
+
+// Route to get all books
+app.get("/books", (req, res) => {
+  try {
+    res.json(booksData);
+  } catch (error) {
+    res.status(500).json({ error: "Internal Server Error" });
+  }
+});
```
FridaKristian, line 58: This is nice looking code.. But I'm not sure the data is collected from the database, I think it might just be directly from the .json-file.

To get data into the database you need to seed it once, this is my code from last week: 
`const seedDataBase = async () => {
  data.forEach(song => {
    new ASong(song).save()
  })
}`
Where I stored my .json-file as the variable data. And my model was called ASong. So in your case you would use "booksData" for the json-file and Books for the model.

And once the data base was seeded, my .get-route looked like:
```try{
const songs = await ASong.find()
if (songs.length>1){
res.json({songs})
}
else {
res.status(404).json({error: "The database is empty, no songs found"})
}
catch (error) {
    res.status(500).json({error: "Something went wrong while loading the songs from the database"})
}```

File: server.js [hunk 3]
```diff
@@ -1,32 +1,117 @@
...
+
+// Create the Author model
+const Author = mongoose.model("Author", authorSchema);
+
+// Route to get all authors
+app.get("/authors", async (req, res) => {
+  try {
+    // Fetch all authors from the database
+    const authors = await Author.find();
```
FridaKristian, line 73: Again, I think you need to seed the Author-database before you can find anything in it.

server.js
```diff
@@ -1,32 +1,117 @@
...
+  );
+
+  if (matchingAuthors.length > 0) {
+    res.json(matchingAuthors);
+  } else {
+    res
+      .status(404)
+      .json({ error: "Authors not found with the specified last name" });
+  }
+});
```
FridaKristian, line 95: I think this might work as long as you have Authors in the database (:

FridaKristian (commented):
Great work this week! I looked at the deployed site and the books-endpoint and a single book worked really well! However I think you need to save the booksdata to the database also

I saw in your comments that you hadn't gotten the authors endpoint to work. I had a similar endpoint, but with artists, and I needed to use the aggregate-method to get it to work. It was the last endpoint I made because it was kind of complicated to figure out the function.. But I think your /authors/:lastName will work if you search through the books by authors. This is my corresponding code for the artists:
```//Route to a specific artist
app.get("/artists/:artist", async (req, res) => {
  const paramArtistName = req.params.artist

  //Regex to make it not case-sensitive (option: i) while searching for any of the artists of a song
  const artistSongs = await ASong.find({ artistName: { $regex : new RegExp(paramArtistName, "i") } });

  if (artistSongs === 0) {
    res.status(404).json("We're sorry, this artist hasn't made any songs in our API")
  }

  res.json(artistSongs)
})
`

All in all I think you're really close to the goal and I hope my code can help you (and not feel overwhelming! Sorry if that's the case ❤️)

PR #447, opened 2023-11-26
Project MongoDB Api - Evelyn Del Carmen

EvelynDelCarmen:
https://mongo-api-evelyn.onrender.com/

### DIFF
--- a/movieModel.js
+++ b/movieModel.js
@@ -1,19 +1,22 @@
+
+// Import the Mongoose library, which helps interact with MongoDB in Node.js
 import mongoose from 'mongoose';
 
 // Define the Mongoose schema for the Movie model
 const movieSchema = new mongoose.Schema({
-    show_id: Number,
-    title: String,
-    director: String,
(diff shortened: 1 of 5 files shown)


movieModel.js
```diff
@@ -0,0 +1,26 @@
+
+// Import the Mongoose library, which helps interact with MongoDB in Node.js
+import mongoose from 'mongoose';
+
+// Define the Mongoose schema for the Movie model
+const movieSchema = new mongoose.Schema({
+    // Define the structure of a movie document in the Movie collection
+    show_id: Number, // Unique identifier for the movie
```
AmandaRam, line 8: MongoDB creates a unique id which means you don't need to specify your own.

AmandaRam (approved):
Really good job! ⭐  The code was easy to read and I like the explanatory comments!
One thing to keep in mind is, as I wrote in the comment, is that MongoDB create a unique id that you can take advantage of further in the code.

PR #449, opened 2023-11-26
mongo api- Elin

Soygirt:
I finished the coding part and got it working on localhost. Had problems figuring out how to deploy and didn't have time to solve them.

### DIFF
--- a/server.js
+++ b/server.js
@@ -1,16 +1,17 @@
 import express from "express";
 import cors from "cors";
 import mongoose from "mongoose";
+import listEndpoints from "express-list-endpoints";
+ import dotenv from "dotenv";
 
-// If you're using one of our datasets, uncomment the appropriate import below
-// to get started!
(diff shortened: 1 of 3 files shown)


server.js
```diff
@@ -23,12 +25,86 @@ const app = express();
...
+    });
+  };
+
+  seedDatabase();
+}*/
 
 // Start defining your routes here
 app.get("/", (req, res) => {
-  res.send("Hello Technigo!");
+  const endpoints = listEndpoints(app);
```
BeckieMorton, line 68: godo to see you used express-list-endpoints. you could even write your code in one line, like this:

app.get("/", (req, res) => {
  res.send(listEndpoints(app));
});

BeckieMorton (commented):
I know this week was difficult Elin and you have a lot going on, but your code works with your local host and looks good and only one more hurdle to jump to get it deployed, so good job!
Please reach out when you need help :-) You can do this, we are nearly at the end!

PR #477, opened 2024-05-10
QUIZ API | Izabel Lind Färnstrand

Izzibizz:
## Onrender link
https://quiz-api-3a1v.onrender.com/


AntonellaMorittu (approved):
Hi Izabel, nice work with your first DB, your code looks very good and it meets all the requirements, all almost all the endpoints are working! Take a look at the `'/question/:id' `endpoint. At the moment it is not working as intended, the problem is with the condition `if (id.length > 0)`. The `findOne` method of Mongoose returns either an object or null if no document matches the query criteria. It does not return an array, so checking the length of the result like if (id.length > 0) will not work as expected.

To fix this, you can check if the `id` variable is not `null` to determine if a question with the given ID exists. Look at this code:

```javascript
app.get('/question/:id', async (req, res) => {
  const question = await Questions.findOne({ id: req.params.id })
  if (question) {
    res.json(question)
  } else {
    res.status(404).json({
      error: 'Could not find a question with this id, the ids range from 0 to 500',
    })
  }
})
``` 
Also generally, remove `id` from the Schema: Mongoose automatically assigns a unique `_id` field to each document, so you don't need to define an `id` field in your schema :)

PR #480, opened 2024-05-11
W14-Mongo API

Maria-Manuela:
## Netlify link
https://project-mongo-api-suco.onrender.comW

### DIFF
--- a/model/Movie.js
+++ b/model/Movie.js
@@ -1,12 +1,12 @@
 import mongoose from "mongoose";
 
-const { Schema, model } = mongoose;
+const { Schema } = mongoose;
 
 const movieSchema = new Schema({
   show_id: {
     type: Number,
     required: true,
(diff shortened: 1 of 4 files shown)


model/Movie.js
```diff
@@ -0,0 +1,43 @@
+import mongoose from "mongoose";
+
+const { Schema, model } = mongoose;
+
+const movieSchema = new Schema({
+  show_id: {
+    type: Number,
+  },
```
AntonellaMorittu, line 8: there's no need to pass an id to your schema, remove `id` as Mongoose automatically assigns a unique `_id` field to each document, so you don't need to define an `id` field  👍

AntonellaMorittu (approved):
Hey Maria-Manuela, your DB works well and the API is documented properly ⭐ The code meets all the specified requirements, being also well-structured and having all the endpoints working!

PR #481, opened 2024-05-11
Project_Mongo_SongAPI-EtnaZR

Caracal23:
## Render link
https://project-mongo-api-qnhd.onrender.com


server.js
```diff
@@ -20,13 +15,95 @@ mongoose.Promise = Promise;
...
+
+app.get("/songs/artist/:artistName", async (req, res) => {
+  const artistName = req.params.artistName;
+  const songsByArtist = await Song.find({
+    artistName,
+  }).exec();
+  if (songsByArtist !== null) {
+    res.json(songsByArtist);
+  } else {
+    res.status(404).send("Information not found, please try again");
```
almaherris, line 105: Same thing as I wrote in the beginning, could be helpful to guide the user on how to search as it's not evident. For example I noticed that it is case sensitive so that could be a good thing to write to help the user :)

almaherris (commented):
Your server.js file is structured and concise and your endpoints are working 😺  I wrote some comments about helpful error messages to enhance user experience and help them work with your API as you intend, as that is something I really tried to implement myself! Good job 😊

server.js
```diff
@@ -20,13 +15,95 @@ mongoose.Promise = Promise;
 const port = process.env.PORT || 8080;
 const app = express();
 
+//Seeding with database
+
+const Song = mongoose.model("Song", {
+  id: Number,
```
AntonellaMorittu, line 21: there's no need to pass an id to your schema, remove `id` as Mongoose automatically assigns a unique `_id` field to each document, so you don't need to define an `id` field :) (also from the data set)

server.js
```diff
@@ -20,13 +15,95 @@ mongoose.Promise = Promise;
...
 // Start defining your routes here
 app.get("/", (req, res) => {
-  res.send("Hello Technigo!");
+  res.send(expressListEndpoints(app));
+});
+
+app.get("/songs", async (req, res) => {
+  const songs = await Song.find();
+  res.json(songs);
+});
```
AntonellaMorittu, line 61: Would be nice to add error handling here, like for example:
```javascript
app.get("/songs", async (req, res) => {
  try {
    const songs = await Song.find();
    res.json(songs);
  } catch (error) {
    res.status(500).json({ error: "Error fetching songs" });
  }
});
```

server.js
AntonellaMorittu: Nice point Alma. The current implementation of the /songs/artist/:artistName endpoint will perform an exact match, which is case-sensitive. To include lowercase and make the search case-insensitive, you can use a regular expression with the $regex operator and the $options set to "i" (for case-insensitive). For example:

```javascript
app.get("/songs/artist/:artistName", async (req, res) => {
  try {
    const artistName = req.params.artistName;
    const songsByArtist = await Song.find({
      artistName: { $regex: new RegExp(artistName, "i") },
    }).exec();

    if (songsByArtist.length > 0) {
      res.json(songsByArtist);
    } else {
      res.status(404).json({ error: "No songs found for this artist" });
    }
  } catch (error) {
    res.status(500).json({ error: "Error fetching songs by artist" });
  }
});
```

AntonellaMorittu (approved):
Hello Etna, you worked well and your DB is up an running! The code is well structured, the API is documented and you met the requirements for this week's assignment. I left some comments in the code, you nailed it! ⭐

PR #486, opened 2024-05-12
W14-Project-mongo-api-Yia

Citronskal:
## Netlify link
https://project-mongo-api-qbfp.onrender.com/ 

## Collaborators
solo

### DIFF
--- a/server.js
+++ b/server.js
@@ -4,6 +4,15 @@ import mongoose from "mongoose";
 import listEndpoints from "express-list-endpoints";
 import topMusicData from "./data/top-music.json";
 
+
+const mongoUrl = process.env.MONGO_URL || "mongodb://localhost/project-mongo";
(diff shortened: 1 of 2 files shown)


server.js
```diff
@@ -1,35 +1,77 @@
...
 mongoose.Promise = Promise;
 
-// Defines the port the app will run on. Defaults to 8080, but can be overridden
-// when starting the server. Example command to overwrite PORT env variable value:
-// PORT=9000 npm start
 const port = process.env.PORT || 8080;
 const app = express();
 
+const Song = mongoose.model("Song", {
+  id: Number,
```
AntonellaMorittu, line 15: You can remove `id` from the Schema: Mongoose automatically assigns a unique `_id` field to each document, so you don't need to define the field in your schema, you can even remove it from the data object

PR #489, opened 2024-05-12
Week 14 – MongoDB API – Nathalie

ohitsnathalie:
## Render link
https://mongo-api-netflixdata.onrender.com


server.js
```diff
@@ -1,35 +1,107 @@
...
+  duration: String,
+  listed_in: String,
+  description: String,
+  type: String,
+})
+
+const NetflixModel = mongoose.model("Netflix", netflixSchema)
+
+// Set Seed
+if (process.env.RESET_DATABASE) {
```
AntonellaMorittu, line 39: did you add the env to the hosting service settings? (Render) The DB is empty at the moment

AntonellaMorittu (changes requested):
Hi Nathalie, congrats deploying your first DB! Your code looks good on github, however I couldn't test the endpoint returning any data correctly, I left a comment as I suspect the issue is the env variable that needs to be passed to Render `RESET_DATABASE=true` in order to seed the db. Please doublecheck, thank you :)

server.js
ohitsnathalie: Hi Antonella,
I checked again and fixed the bug. I saw, that I wrote Netflix instead of NetflixModel. 
It should work now. Thank you :)

AntonellaMorittu (approved):
Nice! Please bear in mind that your data is now having 2 ids, MongoDB assigns automatically an `_id`, so you can remove the `show_id` from your schema and update the code to find the `_id`:

```javascript
const movie = await NetflixModel.findOne({
      _id: Number(id),
      type: "Movie",
    })
```

PR #500, opened 2024-05-16
Booksdata-mongo-api

jingh999:
https://booksdata-mongo-api.onrender.com/
[jingh999]


AntonellaMorittu (approved):
Hi Jing, all looks good, just be aware that MongoDB automatically assigns an `_id` to each data entry, so there's no need to pass a further `id` value. You can update your code based on this to avoid confusion. Thank you! :)

PR #504, opened 2024-12-13
Mongo API

gittebe:
https://project-mongo-api-7ch7.onrender.com/

### DIFF
--- a/server.js
+++ b/server.js
@@ -1,11 +1,16 @@
 import express from "express";
-import bodyParser from "body-parser";
 import cors from "cors";
 import mongoose from "mongoose";
 import booksData from "./data/books.json";
+import dotenv from "dotenv";
 
-const mongoUrl = process.env.MONGO_URL || "mongodb://127.0.0.1:27017/books";
-mongoose.connect(mongoUrl)
+dotenv.config()
+
+// Connecting to MongoDB
+// const mongoUrl = process.env.MONGO_URL || 'mongodb://127.0.0.1:27017/books'
+const mongoURI = process.env.MONGODB_URI
+
+mongoose.connect(mongoURI)
 
   .then(() => {
   console.log('MongoDB successfully connected!')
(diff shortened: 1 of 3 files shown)


server.js
```diff
@@ -1,32 +1,136 @@
...
+  language_code: String,
+  num_pages: Number
+}));
 
 // Defines the port the app will run on. Defaults to 8080, but can be overridden
 // when starting the server. Example command to overwrite PORT env variable value:
 // PORT=9000 npm start
 const port = process.env.PORT || 8080;
 const app = express();
+const listEndpoints = require("express-list-endpoints")
```
HIPPIEKICK, line 39: Can this not be imported like your other packages? 👀

server.js
```diff
@@ -1,32 +1,136 @@
...
+
+if (process.env.RESET_DB)  {
+    const seedDatabase = async () => {
+        await Book.deleteMany({}); 
+        booksData.forEach(book => {
+          new Book(book).save()
+        })
+      }
+    seedDatabase()
+}
```
HIPPIEKICK, line 49: OK to remove once you've seeded your DB

server.js
```diff
@@ -1,32 +1,136 @@
...
+  const books = await Book.find()
+  res.json(books)
+}catch (error) {
+  console.error('Error retrieving books', error);
+    res.status(500).send('Server error');
+}
+});
+
+// Route to get books by author
+app.get('/books/author/:author', async (req, res) => {
```
HIPPIEKICK, line 76: You're still returning books and not authors, so it would be more RESTful to put this as a query param under the /books endpoint

server.js
```diff
@@ -1,32 +1,136 @@
...
+    const books = await Book.find({ language_code: language });
+    if (books.length === 0) {
+      return res.status(404).send('No books found for this language');
+    }
+    res.json(books);
+  } catch (error) {
+    console.error('Error retrieving books by language', error);
+    res.status(500).send('Server error');
+  }
 });
```
HIPPIEKICK, line 134: Same with these filters

HIPPIEKICK (approved):
Since the filters was a stretch goal, I'll allow that you've used separate endpoints for them instead of query params, but please think about it going forward :)

PR #508, opened 2024-12-13
project-mongo-api

zoe-lindqvist:
https://project-mongo-api-pbzi.onrender.com/


server.js
```diff
@@ -1,35 +1,171 @@
...
+const app = express(); // node server.js
+// Defines the port the app will run on
+const port = process.env.PORT || 9000;
 
-const mongoUrl = process.env.MONGO_URL || "mongodb://localhost/project-mongo";
-mongoose.connect(mongoUrl);
-mongoose.Promise = Promise;
+// Add middlewares to enable CORS and JSON body parsing
+app.use(cors());
+app.use(bodyParser.json());
```
HIPPIEKICK, line 18: I think bodyParser is built in to Express nowadays, so it's enough to just connect it like this:
```
app.use(express.json())
```

HIPPIEKICK (approved):
Solid work Zoe! Nice that you tried out some additional HTTP methods :)

PR #509, opened 2024-12-14
Project-Mongo-API - Elina Eriksson Hult

ElinaEH:
https://project-mongo-api-elina.onrender.com


server.js
```diff
@@ -1,35 +1,116 @@
...
+
+    if (books.length === 0) {
+      return res.status(404).send("No books found for this author"); // Handle no results
+    }
+    res.json(books); // Send list of books by the author
+  } catch(error) {
+    console.error("Error retrieving books by author", error); // Log error
+    res.status(500).send("Server error"); // Handle server errors
+  }
 });
```
HIPPIEKICK, line 111: Remember that even though you're filtering on an author, you're still returning _books_. So it would be more RESTful to make this a query param under the /books endpoint. Remember for future!

HIPPIEKICK (approved):
Well done, just remember to use query params for filtering in the future!

PR #512, opened 2024-12-15
Nella: project-mongo-api

nella-x:
https://project-mongo-api-jg3l.onrender.com/


server.js
```diff
@@ -1,35 +1,123 @@
...
+    if (result) {
+      res.json(result);
+    } else {
+      res.status(404).json({ error: "Nominee not found" });
+    }
+  } catch (error) {
+    console.error(`Error fetching nominee: ${nominee}`, error);
+    res.status(500).json({ error: "Internal server error" });
+  }
 });
```
HIPPIEKICK, line 119: Remember to use query params for filters to make it more RESTful. Endpoints should be named after what they return, and you're not returning categories for example, you're returning nominations or golden-globes _filtered_ on a category

HIPPIEKICK (approved):

PR #515, opened 2024-12-15
Maryyy_ux_project-mongo-API

Maryyy-ux:
https://maryyy-ux-project-mongo-api.onrender.com

### DIFF
--- a/server.js
+++ b/server.js
@@ -1,32 +1,121 @@
 import express from "express";
 import cors from "cors";
 import mongoose from "mongoose";
+import { Netflix } from "./models/NetflixModel.js";
+import netflixData from "./data/netflix-titles.json";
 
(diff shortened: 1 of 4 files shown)


server.js
```diff
@@ -1,32 +1,121 @@
...
+  try {
+    const count = await Netflix.countDocuments();
+    res.status(200).json({ totalShows: count });
+  } catch (err) {
+    res.status(500).json({ error: "Failed to fetch stats", details: err.message });
+  }
+});
+
+// Get all shows with pagination
+app.get("/shows", async (req, res) => {
```
HIPPIEKICK, line 83: These three could all just be one endpoint called /shows. Then you can use different query params to filter on specific things

HIPPIEKICK (changes requested):
I get "Service unavailable" on your deployed link, can you please have a look?

PR #517, opened 2024-12-15
Project Mongo API - Jenny A

jempa182:
https://project-mongo-api-npcs.onrender.com/


server.js
```diff
@@ -1,35 +1,142 @@
...
+  }
+  res.json(documentation)
+});
+
+// Get all NPCs with query parameters
+app.get("/npcs", async (req, res) => {
+  const { sorted, gameVersion, lifeStage, category } = req.query;
+
+  try {
+    // Build query object
```
HIPPIEKICK, line 83: Nice! 💯

HIPPIEKICK (changes requested):
Code looks good, but there seems to be something wrong with the deployment. Can you please have a look?

jempa182:
Hi @HIPPIEKICK! I have now checked and apparently I had a Network setting in MongoDB Atlas set so that only my IP adress could view it for some reason.. Very strange though because it worked on Monday. But have updated now and it seems to work now again! Let me know otherwise 😊

server.js
```diff
@@ -1,35 +1,142 @@
...
+    });
+  } catch (error) {
+    res.status(500).json({ success: false, error: error.message });
+  }
+});
+
+// Get single NPC by ID
+app.get("/npcs/:id", async (req, res) => {
+  try {
+    const npc = await Npc.findOne({ id: req.params.id });
```
HIPPIEKICK, line 121: Here you could also use the MongoDB ID

HIPPIEKICK (approved):
Thanks Jenny, got it to work now! ⭐

